PR #2, opened 2019-07-30
Update repo name

rgcostea:


Makefile
```diff
@@ -1,5 +1,5 @@
 NAME = csi-driver
-REPO_NAME ?= hpe/csi-driver
+REPO_NAME ?= hpestorage/csi-driver
```
shivamerla: do we need to update repo name for other images(flexvolume, doryd) as well? Also, in co-deployments(yaml, helm charts etc)

rgcostea: Yes.  I'm working on it.

PR #5, opened 2019-08-02
Docs update

drajen:
Please squash & merge.


README.md
```diff
@@ -40,349 +40,120 @@ A Container Storage Interface ([CSI](https://github.com/container-storage-interf
...
+#### Create a secret with your platform details
+Replace the password string (`YWRtaW4=`) with a base64 encoded version of your password and replace the `backend` with the IP address of the CSP backend and save it as `secret.yaml`:
 
 ```
 apiVersion: v1
 kind: Secret
 metadata:
   name: nimble-secret
-  namespace: default
+  namespace: kube-system
```
shivamerla: This doesn't need to be in kube-system  namespace

USING.md
```diff
@@ -0,0 +1,296 @@
...
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  storageClassName: my-sc-1
+```
+
+**Note:** In most enviornments, there is a default `StorageClass` declared on the cluster. In such a scenario, the `.spec.storageClassName` can be omitted. In addition, the default `StorageClass` need to be annotated with this key: `.metadata.annotations.storageclass.kubernetes.io/is-default-class` set to `"true"`.
```
rgcostea: the default `StorageClass` needs

README.md
drajen: The `default` namespace is non-mandatory for downstream distros. All our bits runs in `kube-system`. Unless we create our own namespace, this is the prettiest option.

PR #11, opened 2019-08-07
change staging mount path to under /var/lib/kubelet

shivamerla:


rgcostea:
Please fix in release-0.3 branch too.

shivamerla:
@rgcostea sure, have pushed new image for v1.0.0-beta, will push v0.3.0-beta as well

PR #13, opened 2019-08-08
Update driver to adhere to CSP updates

rgcostea:


pkg/driver/controller_server.go
```diff
@@ -669,20 +676,36 @@ func (driver *Driver) ControllerPublishVolume(ctx context.Context, request *csi.
...
+	publishInfo, err := storageProvider.PublishVolume(request.VolumeId, node.UUID, requestedAccessProtocol)
 	if err != nil {
 		log.Trace("err: ", err.Error())
 		return nil, status.Error(codes.Internal, "Failed to add ACL via container provider")
 	}
 	log.Tracef("PublishInfo response from CSP: %v", publishInfo)
 
+	// target scope is nimble specific therefore extract it from the volume config
+	var requestedTargetScope = targetScopeGroup
+	if val, ok := existingVolume.Config[targetScope]; ok {
```
shivamerla: So, volume attributes will still have targetScope, but not publishVolume response?

rgcostea: Yes.  I added targetScope to the volume config since this is still needed by CHAPI.  We can eventually do as you suggested and derive this information some other way.

PR #19, opened 2019-08-13
Pod Inline/Ephemeral Volume Support (Mount/Block modes)

suneeth51:
- Added support for POD inline volumes (ephemeral local volumes)
- Supports both Mount and Block access modes
- Added new volume config parameters to be specified in the POD spec:
   1) csi.storage.k8s.io/ephemeral: "true" (This is mandatory parameter for inline volumes)
   2) volumeModeBlock: "true" (This is optional parameter and to be used only for 'Block' mode.
   3) sizeInGiB: "<value_in_GiB>" (This is optional parameter and to be used to specify the size)


pkg/driver/node_server.go
```diff
@@ -486,80 +596,340 @@ func (driver *Driver) NodePublishVolume(ctx context.Context, request *csi.NodePu
...
+	if err := driver.UpdateDB(dbKey, volume); err != nil {
+		return err
+	}
+
+	// Get Node Info
+	nodeID, err := driver.nodeGetInfo()
+	if err != nil {
+		log.Tracef("Failed to get node info, err: %s", err.Error())
+		return status.Error(codes.Internal,
+			fmt.Sprintf("Failed to node publish ephemeral volume %s due to error: %s", ephemeralVolName, err.Error()))
```
shivamerla: don't we need a rollback if anything else fails after volume is created? I think nodeUnpublish will not be called to cleanup volume if nodePublish fails due to some error at some point.

examples/kubernetes/clone-from-pvc-resize.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: clone-of-pvc-1-resized
+  namespace: kube-system
```
rgcostea: drop the namespace

examples/kubernetes/clone-from-pvc.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: clone-of-pvc-1
+  namespace: kube-system
```
rgcostea: drop the namespace

examples/kubernetes/pod-inline-block.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: my-pod-inline-block-1
+  namespace: kube-system
```
rgcostea: drop the namespace

examples/kubernetes/pod-inline.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: my-pod-inline-mount-1
+  namespace: kube-system
```
rgcostea: drop the namespace

pkg/driver/constants.go
```diff
@@ -49,5 +52,23 @@ const (
 	Pending = "PENDING"
 
 	// common properties
-	descriptionKey = "description"
+	descriptionKey     = "description"
+	destroyOnDeleteKey = "destroyOnDelete"
```
rgcostea: This is actually specific to the CSP implementation.  A different CSP might choose to destroy on delete by default.  The nimble one does not currently.  We should probably formalize this in the CSP spec.

pkg/driver/controller_server.go
```diff
@@ -241,25 +218,65 @@ func (driver *Driver) CreateVolume(ctx context.Context, request *csi.CreateVolum
...
+
+// nolint: gocyclo
+func (driver *Driver) createVolume(
+	name string,
+	size int64,
+	volumeCapabilities []*csi.VolumeCapability,
+	secrets map[string]string,
+	volumeContentSource *csi.VolumeContentSource,
+	requestParameters map[string]string,
+	createParameters map[string]string) (*csi.Volume, error) {
```
rgcostea: The only difference between the requestParameters and the createParameters is that the createParameters apply the overrides from the PVC annotations.  I don't think we need to pass in both.  Just use the createParameters.

pkg/driver/controller_server.go
```diff
@@ -671,64 +737,63 @@ func (driver *Driver) ControllerPublishVolume(ctx context.Context, request *csi.
...
 
 	// Add ACL to the volume based on the requested Node ID
-	publishInfo, err := storageProvider.PublishVolume(request.VolumeId, node.UUID, requestedAccessProtocol)
+	publishInfo, err := storageProvider.PublishVolume(volume.ID, node.UUID, requestedAccessProtocol)
 	if err != nil {
-		log.Trace("err: ", err.Error())
-		return nil, status.Error(codes.Internal, "Failed to add ACL via container provider")
+		log.Errorf("Failed to publish volume %s, err: %s", volume.ID, err.Error())
+		return nil, status.Error(codes.Internal,
+			fmt.Sprintf("Failed to add ACL to volume %s for node %v via container provider", volume.ID, node))
```
rgcostea: let's replace `container provider` with CSP

pkg/driver/node_server.go
rgcostea: Agreed.  We should rollback if anything fails after creation.

pkg/flavor/kubernetes/flavor.go
```diff
@@ -406,3 +407,79 @@ func (flavor *Flavor) getNodeInfoByUUID(uuid string) (*crd_v1.HPENodeInfo, error
...
+			//if vol.Name == volName {
+			csiSource := vol.VolumeSource.CSI
+			if csiSource == nil {
+				return nil, fmt.Errorf("CSI volume source is nil")
+			}
+
+			// No secrets are configured
+			if csiSource.NodePublishSecretRef == nil {
+				log.Trace("No secrets are configured")
+				return secrets, nil
```
rgcostea: Shouldn't this return an error since we require the secrets?

pkg/driver/node_server.go
raunakkumar: Should we look into using the command chains for create / execute /rollback ? https://github.com/hpe-storage/common-host-libs/blob/master/chain/chain.go

suneeth51: > don't we need a rollback if anything else fails after volume is created? I think nodeUnpublish will not be called to cleanup volume if nodePublish fails due to some error at some point.

@shivamerla We don't need cleanup as it will be retried by the k8s(external-provisioner) whenever there is a failure of NodePublish(). Also, all CSI calls are implemented to be idempotent, so it will continued from where it is left of to complete the operation.

shivamerla: @suneeth51 thats true, it keeps retrying to attach the volume, but it cannot go through for some reason(say multipath is not responding etc), they might end up deleting the pod. But since there is no PVC/PV associated with it, volume will be lying around. After this there is no way to delete the volume i think?  Let me know if i am missing something..

examples/kubernetes/clone-from-pvc-resize.yaml
suneeth51: Agreed, will remove it.

pkg/driver/constants.go
suneeth51: Agreed, the non-nimble CSP should still be able to destroy with/without this metadata key.

Alternatively, why don't we implement the CSP's Delete API to destroy the volume by default unless explicitly requested by the user via SC to retain the volume by just offline it...

pkg/flavor/kubernetes/flavor.go
suneeth51: Agreed. I will fix this to return appropriate error when it is unspecified.

pkg/driver/node_server.go
suneeth51: Agreed. I will update the logic to handle the rollback/cleanup to remove the volume from the storage.

README.md
```diff
@@ -52,6 +52,9 @@ Kubernetes 1.14
  * `--allow-privileged` flag must be set to true for the API server
  * `--feature-gates=ExpandCSIVolumes=true,ExpandInUsePersistentVolumes=true` feature gate flags must be set to true for both the API server and kubelet
```
rgcostea: Copy this to the 1.15 section and add " for resize support"

README.md
```diff
@@ -52,6 +52,9 @@ Kubernetes 1.14
...
 
+Kubernetes 1.15
+ * `--feature-gates=CSIInlineVolume=true,VolumePVCDataSource=true` feature gate flags must be set to true for both the API server and kubelet
```
rgcostea: Add " for ephemeral inline volume and Volume cloning" . Or separate them into 2 different bullet points.

pkg/driver/driver.go
```diff
@@ -330,6 +330,65 @@ func (driver *Driver) GetVolumeByID(id string, secrets map[string]string) (*mode
...
+		}
+
+		// check if the volume exists
+		volume, err = storageProvider.GetVolumeByName(name)
+		if err != nil {
+			log.Error("err: ", err.Error())
+			return nil, status.Error(codes.Internal, fmt.Sprintf("Error while attempting to get volume %s", name))
+		}
+	} else {
+		log.Tracef("Secrets not provided. Checking all known storage providers.")
```
rgcostea: I'm not a fan of doing this by name.  The ID is a UUID.  The name is not.  We should probably only support the case where the secrets are provided.

pkg/driver/node_server.go
```diff
@@ -753,8 +774,111 @@ func (driver *Driver) nodePublishVolume(
...
+			volumeName,
+			secrets,
+			stagingTargetPath,
+			targetPath,
+		)
+		if err != nil {
+			if try < maxTries {
+				try++
+				log.Tracef("Retry attempt %d unsuccessful. Pending retries %d", try, maxTries-try)
+				time.Sleep(time.Duration(try) * time.Second)
```
rgcostea: awesome!

pkg/driver/node_server.go
```diff
@@ -810,23 +934,16 @@ func (driver *Driver) nodePublishEphemeralVolume(
...
-		if err != nil {
-			return err
-		}
-		size = int64(value * GiBToByte) // Convert GiB to Bytes
+	// Get the volume size from the volume context if specified
+	sizeStr := volumeContext[sizeKey]
+	if sizeStr != "" {
+		volSize := resource.MustParse(sizeStr)
+		sizeInBytes = volSize.Value()
+		log.Tracef("Ephemeral volume %s requested with size %s (%v)", sizeStr, sizeInBytes)
```
rgcostea: Missing first parameter (volume).

pkg/driver/driver.go
suneeth51: Agreed. Will change to lookup with specific provider for the given secrets.

pkg/driver/node_server.go
suneeth51: My bad.. I will fix this.

pkg/driver/controller_server.go
suneeth51: Agreed. I will change it accordingly.

PR #39, opened 2019-08-29
use nfs-ganesha image instead of volume-nfs

shivamerla:


pkg/flavor/kubernetes/nfs.go
```diff
@@ -23,13 +23,13 @@ import (
 const (
 	nfsPrefix    = "hpe-nfs-"
 	nfsNamespace = "hpe-nfs"
-	nfsImage     = "gcr.io/google_containers/volume-nfs:0.8"
+	nfsImage     = "nimblestorage/nfs-ganesha"
```
rgcostea: Do we have a tag for this?

shivamerla: sure George, will update the version tag.

PR #44, opened 2019-09-16
vst volumes are not getting logged out

raunakkumar:
- csp returns target_scope

### DIFF
--- a/pkg/driver/controller_server.go
+++ b/pkg/driver/controller_server.go
@@ -761,6 +761,13 @@ func (driver *Driver) controllerPublishVolume(
 		return nil, err
 	}
 
+	volumeConfig := make(map[string]interface{})
+	for k, v := range volume.Config {
+		// convert volume config from CSP to camelCase before parsing it
+		volumeConfig[util.ToCamelCase(k)] = v
+	}
+	log.Tracef("volume config is %+v", volumeConfig)
+
 	// Decode and check if the node is configured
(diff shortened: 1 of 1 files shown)


pkg/driver/constants.go
```diff
@@ -37,7 +37,7 @@ const (
 	serialNumberKey     = "serialNumber"
 	accessProtocolKey   = "accessProtocol"
 	targetNameKey       = "targetName"
-	targetScopeKey      = "targetScope"
+	targetScopeKey      = "target_scope"
```
rgcostea: CSP no longer returns target_scope in the publish response.  We should be able to determine this by reading the volume attribute named target_scope.  JSON is always reported in snake_case.  However, we should still map that to camelCase and leave this alone.  What is the actual problem?

raunakkumar: This is what gets returned as part of volume.Config
```volume config is map[clone_of: dedupe_enabled:false destroy_on_delete:true encrypted:false folder: limit_iops:76800 limit_mbps:-1 performance_policy:SQL Server pool:default sync_on_detach:false target_scope:volume thick:false]" file="controller_server.go:824"```

Do you think something has changed while unmarshalling the object?

rgcostea: Ah.  This is because there is no VolumeConfig struct defined in types.go specific to Nimble.  If there were, we could add a line like this:
	TargetScope   string                 `json:"target_scope,omitempty"`
Which would unmarshall the JSON response to TargetScope?

PR #47, opened 2019-10-08
Improved error log messages

suneeth51:


suneeth51:
> Should we log this on other functions too? Publish/Unpublish, etc.
Errors were always been logged, I just fixed the verbiage here.

PR #55, opened 2019-10-23
Support multiple discovery IPs from CSP

rgcostea:


pkg/driver/node_server.go
```diff
@@ -382,13 +383,17 @@ func (driver *Driver) setupDevice(publishContext map[string]string) (*model.Devi
 	defer log.Trace("<<<<< setupDevice")
 
 	// TODO: Enhance CHAPI to work with a PublishInfo object rather than a volume
+
+	discoveryIps := strings.Split(publishContext[discoveryIPsKey], ",")
```
suneeth51: May need to check for empty list?

Issue #51, opened 2019-10-21
The `hpe-csi-node` DaemonSet is way too invasive and does not work on RHEL/CentOS/Fedora Atomic.


Logibox:
RHEL/CentOS/Fedora Atomic has a immutable filesystem so you can't inject systemd units or install packages. Why isn't `iscsid` and `multipathd` provided by the container?

shivamerla:
@Logibox So far we haven't verified our drivers on Atomic host or CoreOS. In CoreOS both iSCSI and multipath packages are already pre-installed with base image. I believe with RedHat Atomic Host 7.2.2 both are pre-packaged as well. We will handle these and bypass package installation from our systemd scripts. We will also look into allowing users to control package installation by the driver.

shivamerla:
@Logibox we have added env param DISABLE_NODE_CONFORMANCE, when passed to node-plugin it will by-pass node conformance checks. Once we add support for CoreOS/Atomic we will update those YAML/deployment files accordingly.

PR #58, opened 2019-10-24
Run iscsid and multipathd inside the container

Logibox:
Proposed fix for #51, though it's pretty hacky at the moment it should prove the concept. This requires https://github.com/hpe-storage/co-deployments/pull/72 to work.

Rather that doing very suspect modifications to the host, just run daemons with the container itself. This has the following advantages:
* More portable - all the host is required to have is the correct kernel modules loaded. So distribution specific commands (like in `hpe-storage-node.sh` for example) aren't required.
* Less risky - no low level modifications to the host are required, which means this driver can be cleanly uninstalled (also make upgrades easier).
* More control over setting - although I've not shown it here, rather than editing the host `/etc/iscsi/iscsid.conf`, you can just embed one in the container. Or better yet - put it in a `ConfigMap`.


cmd/csi-driver/conform/entrypoint.sh
```diff
@@ -7,30 +7,14 @@ set -xe
 for arg in "$@"
 do
     if [ "$arg" = "--node-service" ]; then
-        nodeservice=true
+        if [ ! -z "$iSCSI_INITIATOR_NAME" ]; then
+            echo "InitiatorName=${iSCSI_INITIATOR_NAME}" > /etc/iscsi/initiatorname.iscsi
+        fi
+        iscsid -f &
+        multipathd -d &
```
Logibox: This could be improved by using `supervisord`.

shivamerla:
@Logibox We had tried this as first attempt when developing the driver and hit into multiple issues. They are actually well summarized by docker team here: https://engineering.docker.com/2019/07/road-to-containing-iscsi/ 

I am sure if you try to test this with CentOS and Ubuntu images with volume mounts, you will definitely run into these issues.

Logibox:
Thanks for the link - interesting read. However I would argue that this setup isn't quiet what's describe in any of the options in the article, though it is similar to option 1.

This approach doesn't require a insmod, instead relies on the stability of the netlink API between the user space tool provided by the container and the kernel module on the host. Admittedly I don't have a source whether that API is stable or not, but I'll look that up.

Also this isn't really true:
> Also, on systemd based distros, systemd socket activation on the host keeps the socket busy, even if the service is not running. So the entire iscsi host package has to be removed from the host. ( eg. `yum remove iscsi-initiator-tools` or `apt-get remove open-iscsi`).

You can just disable the systemd socket:
```
systemctl disable iscsid.socket
systemctl stop iscsid.socket
```
Which is a bit less invasive than uninstalling packages. Also this is disabled by default in CentOS 8 and Fedora 30.

Option 3 looks to be a good solution, but in solution given the host volume isn't modified. The main issue I have with the current approach that you have right now is the the modifications of the host OS. Is there any way the driver container could move to that model where the host fs is a read-only volume?

Logibox:
Just checked - the netlink API is stable https://github.com/open-iscsi/open-iscsi/issues/177

rgcostea:
If we disable the systemd socket, then you've effectively disabled iscsi from the host.  It's only available for containers.  What if you want volume mounts directly to your hosts as well?

Logibox:
That's a fair point - there's also the issue of potentially multiple instances of `iscsid` running if another CSI driver tries this.
But is mount volume directly to a k8s node a valid use case? I can't think of a reason I would do that.

rgcostea:
iSCSI boot would be a use case.

drajen:
A non-starter for `iscsid` in a container when we (ping @e4jet) looked at it years ago for the managed DVP is that there's no safe way to upgrade the container. A restart will sever every connection and wreak havoc. This could probably be worked around somehow with clever configuration to be as non-disruptive as possible to user applications but that was too much of a stumble and we chose the safe path of mapping the host `iscsiadm` into the container. A more favorable design in this case would be to run `iscsid` and `multipathd` in a `DaemonSet` that you'll rarely ever touch and have it running almost statically. That would require a pretty hefty redesign on our part though I suppose.

Another approach would be similar to how Istio injects sidecars into your workloads that uses Istio. If your workload is using iSCSI/multipathing, we would simply inject the sidecar with it's own `iscscid` and `multipathd`. For this to work though, I think `iscsid` needs to be namespace aware. Not sure how that project is coming along and not sure how invasive users want their block storage to be.

PR #62, opened 2019-10-31
Check if device exists on the host during NodeStage

suneeth51:


cmd/csi-driver/diag/hpe-logcollector.sh
```diff
@@ -78,9 +78,7 @@ mkdir -p $directory
 nimble_ftp_site="ftp.nimblestorage.com"
 nimble_ftp_folder="upload"
 nimble_ftp_user="anonymous"
-nimble_ftp_password="hpestoragelogupload@nimblestorage.com"
-
-
+nimble_ftp_password="foo@bar"
```
rgcostea: Does this work?  Or does it check for ".com" or ".org" on the end?

PR #79, opened 2019-11-19
add timeout to some diag commands

raunakkumar:
- multipath and other commands may hang because of the system state. introduce timeout to collect the diag if these hang


raunakkumar:
yup that was next on to shorten this commands to work with kubectl

PR #82, opened 2019-11-27
csi-driver : diagnostic to work with kubectl

raunakkumar:
* Problem:
- diagnostic log collection was not working using kubectl
* Implementation:
- include hpe script to container path
- deprecate use of ftp and kubectl commands within the diagnostic script
- include a hpe-kubectl-diagnostic.sh for users to invoke the logs
from any node which has access to kubectl of the cluster
* Testing:

./hpe-kubectl-diagnostic.sh 1234
Please enter a valid node name

./hpe-kubectl-diagnostic.sh rkumar-k8s-cl1-2
======================================
Collecting the Diagnostic Information
======================================
Diagnostic dump file created at /var/log/hpe-storage-logs-rkumar-k8s-cl1-2-20191127_185857.tar.gz on host rkumar-k8s-cl1-2
Complete
====================================

./hpe-kubectl-diagnostic.sh
======================================
Collecting the Diagnostic Information
======================================
Diagnostic dump file created at /var/log/hpe-storage-logs-rkumar-k8s-cl1-2-20191127_185903.tar.gz on host rkumar-k8s-cl1-2
Complete
====================================
======================================
Collecting the Diagnostic Information
======================================
Diagnostic dump file created at /var/log/hpe-storage-logs-rkumar-k8s-cl1-3-20191127_185904.tar.gz on host rkumar-k8s-cl1-3
Complete
====================================

* Bug: https://nimblejira.nimblestorage.com/browse/CON-469

### DIFF
--- a/cmd/csi-driver/Dockerfile
+++ b/cmd/csi-driver/Dockerfile
@@ -33,7 +33,4 @@ ADD [ "csi-driver", "/bin/" ]
 ADD [ "conform/entrypoint.sh", "/entrypoint.sh" ]
 RUN [ "chmod", "+x", "/entrypoint.sh" ]
 
-# add path to hpe scripts
-ENV PATH="/usr_local/local/bin:${PATH}"
-
 ENTRYPOINT [ "/entrypoint.sh" ]
--- a/cmd/csi-driver/Dockerfile
+++ b/cmd/csi-driver/Dockerfile
@@ -33,4 +33,7 @@ ADD [ "csi-driver", "/bin/" ]
 ADD [ "conform/entrypoint.sh", "/entrypoint.sh" ]
 RUN [ "chmod", "+x", "/entrypoint.sh" ]
 
+# add path to hpe scripts
+ENV PATH="/usr_local/local/bin:${PATH}"
+
 ENTRYPOINT [ "/entrypoint.sh" ]
--- a/hpe-kubectl-diagnostic.sh
+++ b/hpe-kubectl-diagnostic.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+# (c) Copyright 2019 Hewlett Packard Enterprise Development LP
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+# http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


cmd/csi-driver/Dockerfile
```diff
@@ -33,4 +33,7 @@ ADD [ "csi-driver", "/bin/" ]
 ADD [ "conform/entrypoint.sh", "/entrypoint.sh" ]
 RUN [ "chmod", "+x", "/entrypoint.sh" ]
 
+# add path to hpe scripts
+ENV PATH="/usr_local/local/bin:${PATH}"
```
shivamerla: since we are using kubectl to exec and collect logs we don't need to put the script on host path. It can be within container /usr/local/bin etc and not hosts user_local path.

raunakkumar: since we have mapped /usr/local to /usr_local/local inside the container the scripts are not in path. I get the following error without setting this path

root@rkumar-k8s-cl1-1:~# ./hpe-kubectl-diagnostic.sh rkumar-k8s-cl1-2
OCI runtime exec failed: exec failed: container_linux.go:346: starting container process caused "exec: \"hpe-logcollector.sh\": executable file not found in $PATH": unknown
command terminated with exit code 126

shivamerla: Yes, if we change cp from /usr_local to container path /usr/local then it will be part of PATH variable.

hpe-kubectl-diagnostic.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
shivamerla: I think we can still call this hpe-log-collector.sh as other one is internal to us.

cmd/csi-driver/Dockerfile
raunakkumar: yup works with the change. Thanks

hpe-kubectl-diagnostic.sh
```diff
@@ -0,0 +1,66 @@
...
+	for i in `kubectl get pods -n kube-system --selector=app=hpe-csi-node -o json -o jsonpath='{.items..metadata.name}'` ;
+	do
+		kubectl exec -it $i -c hpe-csi-driver -n kube-system -- hpe-logcollector.sh;
+	done;
+fi
+
+}
+
+display_usage() {
+echo "Diagnostic Script to collect HPE Storage logs using kubectl"
```
shivamerla: if possible, please add an option like --node-name and cli parsing logic to be more standard. if --node-name is not present we can document saying default is --all.

PR #83, opened 2019-12-03
csi-driver: add env param to disable node conformance checks

shivamerla:
* Problem:
  * In some of the OS distributions(Atomic/CoreOS), we cannot inject systemd files or install
  * packages as OS is immutable.
* Implementation:
  * Add env param DISABLE_NODE_CONFORMANCE which user can specify to bypass package checks and installation
* Testing: tested with env param set and verified no conformance/systemd file are copied.
* Review: gcostea, rkumar
https://github.com/hpe-storage/csi-driver/issues/51

### DIFF
--- a/cmd/csi-driver/conform/entrypoint.sh
+++ b/cmd/csi-driver/conform/entrypoint.sh
@@ -12,17 +12,21 @@ do
 done
 
 if [ "$nodeservice" = true ]; then
-    # Copy HPE Storage Node Conformance checks and conf in place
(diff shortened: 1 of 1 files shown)


shivamerla:
> lgtm. maybe just add it to the documentation wherever we claim support
Yes, will add this param to helm charts to allow disabling. But we can do that once we support CoreOS/Atomic. I have added env to allow users to try it out on CoreOS with YAML changes.

PR #85, opened 2019-12-04
csi : return retry able error code for csi-attacher

raunakkumar:
* Problem:
- codes.Internal is treated as final error with csi attacher
* Implementation:
- modify the codes.Internal for GetVolume and GetVolumeID to be
of type codes.Unavailable to that csi attacher retries in case
of controllerUnpublish
* Testing:
* Bug: https://nimblejira.nimblestorage.com/browse/CON-573

### DIFF
--- a/pkg/driver/controller_server.go
+++ b/pkg/driver/controller_server.go
@@ -369,7 +369,7 @@ func (driver *Driver) createVolume(
 	storageProvider, err := driver.GetStorageProvider(secrets)
 	if err != nil {
 		log.Error("err: ", err.Error())
-		return nil, status.Error(codes.Internal, fmt.Sprintf("Failed to get storage provider from secrets, %s", err.Error()))
+		return nil, status.Error(codes.Unavailable, fmt.Sprintf("Failed to get storage provider from secrets, %s", err.Error()))
 	}
 
 	// TODO: check for version compatibility for new features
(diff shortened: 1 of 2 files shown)


pkg/driver/controller_server.go
```diff
@@ -930,7 +930,10 @@ func (driver *Driver) controllerUnpublishVolume(volumeID string, nodeID string,
 	storageProvider, err := driver.GetStorageProvider(secrets)
 	if err != nil {
 		log.Error("err: ", err.Error())
-		return status.Error(codes.Internal,
```
shivamerla: does this apply to ControllerPublish as well? I see we use `Internal` error code with multiple failures.

raunakkumar: Ya it does https://github.com/kubernetes-csi/external-attacher/blob/ad7d376eb4ac985860e37903590ce6a17b340b06/pkg/attacher/attacher.go#L72 
but I looked at those error codes and those are actual volume creation errors and we shouldn't retry. I'll double check if we missed any

shivamerla: creation? this is during ControllerPublish right? I was checking calls like GetStorageProvider(), GetNodeContext() and SetNodeContext() which returns `Internal` error code. May be on certain http status codes we should indicate caller to derive at error code which can be retried or always return `Unavailable`, @george?

raunakkumar: Ya, I looked at volume creation ones. GetStorageProvider(), GetNodeContext() and SetNodeContext() could be retried.

PR #89, opened 2019-12-05
csi: update hpenodeinfo if exists

raunakkumar:
* Problem:
- during node hard reset if the CRD still
exists the nodeID of the node could change and may conflict
with existing hpenodeinfo with same name
- the networks,iqns and wwpns of host could change without
change in nodename or node UUID
* Implementation:
- lookup by uuid followed by name before claiming no hpenodeinfo
object present
- check if iqns,wwpns or networks need to be updated for
existing crd and update if needed
* Testing:
- tested by modifying nodeID in node.gob to a different value and recreated csi pod bringup updates the existing CRD
* Bug: https://nimblejira.nimblestorage.com/browse/CON-574

### DIFF
--- a/pkg/flavor/kubernetes/flavor.go
+++ b/pkg/flavor/kubernetes/flavor.go
@@ -7,6 +7,7 @@ import (
 	b64 "encoding/base64"
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/flavor/kubernetes/flavor.go
```diff
@@ -139,31 +140,46 @@ func (flavor *Flavor) LoadNodeInfo(node *model.Node) (string, error) {
...
-		var wwpns []string
-		for i := 0; i < len(node.Wwpns); i++ {
-			wwpns = append(wwpns, *node.Wwpns[i])
+		// make sure the nodeInfo has updated information from the host
+		if !reflect.DeepEqual(nodeInfo.Spec.IQNs, getIqnsFromNode(node)) ||
+			!reflect.DeepEqual(nodeInfo.Spec.WWPNs, getWwpnsFromNode(node)) ||
+			!reflect.DeepEqual(nodeInfo.Spec.Networks, getNetworksFromNode(node)) {
+			nodeInfo.Spec.IQNs = getIqnsFromNode(node)
+			nodeInfo.Spec.Networks = getNetworksFromNode(node)
+			nodeInfo.Spec.WWPNs = getWwpnsFromNode(node)
```
rgcostea: I think it looks cleaner to do this in 3 separate if blocks...

iqnsFromNode := getIqnsFromNode(node)
if !reflect.DeepEqual(nodeInfo.Spec.IQNs, iqnsFromNode) {
   node.Spec.IQNs = iqnsFromNode
}
...

PR #94, opened 2019-12-11
update TAG to be branchname  for future branches

raunakkumar:


raunakkumar:
reverting to use tag and not branch name

PR #97, opened 2019-12-19
Support multiple targets per volume(3par)

shivamerla:


shivamerla:
> lgtm. Are you working on the CSP changes too?

yes, will make those changes and perform end to end tests. So far i was able to verify by converting single entry into list.

PR #104, opened 2020-01-29
csi: move to ubi and use chroot mechnism for host utilities

shivamerla:
Problem:
        Need uniform image to support across OCP and other K8s flavors
        Need to minimize dependency of packages within container from
        underlying host libs(eg, iscsi, systemd deps between RHEL7 vs
        RHEL 8 etc)
    Implementation:
        Switched CSI driver image from debian based to UBI 7 minimal
        Removed un-necessary mount paths with deployment and use chroot instead
        with host root fs mounted within container
        Copy systemd service files to /etc/ paths rather than /lib.
        Handle multipath.conf changes during startup and avoid multipath.conf mount(RKE limitation)
    Testing: tested on 1.15 and 1.16 and verified pod creation, volume mounts work fine. Planning to run
    complete integration suite before merge.
    Review: gcostea, rkumar, sbyadarahalli
    Bug: https://nimblejira.nimblestorage.com/browse/NLT-
    Signed-off-by: Shiva Krishna, Merla shivakrishna.merla@hpe.com

### DIFF
--- a/cmd/csi-driver/Dockerfile
+++ b/cmd/csi-driver/Dockerfile
@@ -34,8 +34,7 @@ RUN ln -s /chroot/chroot-host-wrapper.sh /chroot/blkid \
     && ln -s /chroot/chroot-host-wrapper.sh /chroot/multipathd \
     && ln -s /chroot/chroot-host-wrapper.sh /chroot/umount \
     && ln -s /chroot/chroot-host-wrapper.sh /chroot/ip \
-    && ln -s /chroot/chroot-host-wrapper.sh /chroot/dmidecode \
-    && ln -s /chroot/chroot-host-wrapper.sh /chroot/dmsetup
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/dmidecode
 
 ENV PATH="/chroot:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 
--- a/cmd/csi-driver/Dockerfile
+++ b/cmd/csi-driver/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi7/ubi-minimal:7.7-138
+FROM registry.access.redhat.com/ubi7/ubi-minimal:7.7-238
 
 LABEL name="HPE CSI Driver for Kubernetes" \
       maintainer="HPE Storage" \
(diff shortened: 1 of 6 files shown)


cmd/csi-driver/Dockerfile
```diff
@@ -1,17 +1,45 @@
...
+      io.k8s.description="The HPE CSI Driver for Kubernetes enables container orchestrators, such as Kubernetes and OpenShift, to manage the life-cycle of persistent storage." \
+      io.openshift.tags=hpe,csi,hpe-csi-driver
+
+WORKDIR /root
+COPY /LICENSE /licenses/
+
+RUN mkdir /chroot
+ADD chroot-host-wrapper.sh /chroot
+RUN chmod 777 /chroot/chroot-host-wrapper.sh
+RUN    ln -s /chroot/chroot-host-wrapper.sh /chroot/blkid \
```
raunakkumar: few other command I found in common-host-libs/linux 
udevadm
lsb_release
uname

cmd/csi-driver/Dockerfile
```diff
@@ -1,17 +1,45 @@
...
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/btrfs \
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/fsck \
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/mount \
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/multipath \
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/multipathd \
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/umount \
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/ip \
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/dmidecode \
+    && ln -s /chroot/chroot-host-wrapper.sh /chroot/systemctl \
+     && ln -s /chroot/chroot-host-wrapper.sh /chroot/dmsetup
```
rgcostea: nit: extra space

cmd/csi-driver/Dockerfile
shivamerla: these are already part of UBI and works well within container as we mount /etc/os-release file paths. Udevadm is not yet used with CSI.

cmd/csi-driver/Dockerfile
shivamerla: fixed..

PR #109, opened 2020-02-19
csi-driver: misc fixes related to nodeinfo updates

shivamerla:
* Problem:
  * nodeinfo resources are not cleaned up as log.Fatalf was used before unload.
  * ubi-init images doesn't receive signal on kill
  * on node-name re-use, stale nodeinfo resources are re-used as well
* Implementation:
  * change log.Fatalf to Infof so we just log it but not exit.
  * change base images to ubi-minimal which works well in all platforms.
  * update nodeinfo properties on node name re-use and stale hpenodeinfo objects.
* Testing: tested scenarios of creation/deletion/IQN updates. node scale up/down by Fred.

### DIFF
--- a/cmd/csi-driver/csi-driver.go
+++ b/cmd/csi-driver/csi-driver.go
@@ -148,7 +148,7 @@ func csiCliHandler(cmd *cobra.Command) error {
 		syscall.SIGTERM)
 
 	s := <-stop
-	log.Fatalf("Exiting due to signal [%v] notification for pid [%d]", s.String(), pid)
+	log.Infof("Exiting due to signal [%v] notification for pid [%d]", s.String(), pid)
 	d.Stop(nodeService)
 	log.Infof("Stopped [%d]", pid)
 	return nil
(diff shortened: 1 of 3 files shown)


cmd/csi-driver/csi-driver.go
```diff
@@ -148,7 +148,7 @@ func csiCliHandler(cmd *cobra.Command) error {
 		syscall.SIGTERM)
 
 	s := <-stop
-	log.Fatalf("Exiting due to signal [%v] notification for pid [%d]", s.String(), pid)
+	log.Infof("Exiting due to signal [%v] notification for pid [%d]", s.String(), pid)
```
rgcostea: Does log.Fatalf exit?  Is that expected?

pkg/flavor/kubernetes/flavor.go
```diff
@@ -134,29 +134,50 @@ func (flavor *Flavor) LoadNodeInfo(node *model.Node) (string, error) {
...
 			nodeInfo.Spec.IQNs = iqnsFromNode
 			updateNodeRequired = true
 		}
+		// update node network information on mismatch
 		networksFromNode := getNetworksFromNode(node)
 		if !reflect.DeepEqual(nodeInfo.Spec.Networks, networksFromNode) {
 			nodeInfo.Spec.Networks = networksFromNode
 			updateNodeRequired = true
 		}
+		// update node FC port WWPN's on mismatch
```
rgcostea: nit: WWPNs

shivamerla: fixed..

PR #110, opened 2020-02-19
csi-driver: misc fixes related to nodeinfo updates (#109)

shivamerla:
* csi-driver: misc fixes related to nodeinfo updates
* Problem:
  * nodeinfo resources are not cleaned up as log.Fatalf was used before unload.
  * on node-name re-use, stale nodeinfo resources are re-used as well
* Implementation:
  * change log.Fatalf to Infof so we just log it but not exit.
  * update nodeinfo properties on node name re-use and stale hpenodeinfo objects.
* Testing: tested scenarios of creation/deletion/IQN updates. node scale up/down by Fred.

### DIFF
--- a/cmd/csi-driver/csi-driver.go
+++ b/cmd/csi-driver/csi-driver.go
@@ -148,7 +148,7 @@ func csiCliHandler(cmd *cobra.Command) error {
 		syscall.SIGTERM)
 
 	s := <-stop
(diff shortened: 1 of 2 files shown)


shivamerla:
> lgtm.
> Are we spinning another build of release-1.0?

Yes, we are spinning 1.0.1

PR #111, opened 2020-02-26
Fixed bug CON-693

vidhutsingh:
Problem: 
PVC creation should error out if filesystem type of child In storage class definition is different from parent PVC

Fix:
Added a logic to fetch the parent volume filesystem type and compare it with the requested pvc filesystem, error out if they are different.

### DIFF
--- a/pkg/driver/controller_server.go
+++ b/pkg/driver/controller_server.go
@@ -430,9 +430,7 @@ func (driver *Driver) createVolume(
 
 	// Create volume using pre-populated content if requested
 	if volumeContentSource != nil {
-		log.Tracef("Request with volumeContentSource: %+v", volumeContentSource)
-
-		// Verify source snap if specified
+		log.Tracef("Request with volumeContentSource: %+v", volumeContentSource) // Verify source snap if specified
 		if volumeContentSource.GetSnapshot() != nil {
 			// Check if the driver supports SNAPSHOT service
 			if !driver.IsSupportedControllerCapability(csi.ControllerServiceCapability_RPC_CREATE_DELETE_SNAPSHOT) {
@@ -457,6 +455,32 @@ func (driver *Driver) createVolume(
 				return nil, status.Error(codes.Internal, "Parent volume ID is missing in the snapshot response")
 			}
 
+			// Get the parent volume from the snapshot
+			existingParentVolume, err := storageProvider.GetVolume(existingSnap.VolumeID)
+			if err != nil {
+				log.Error("err: ", err.Error())
+				return nil,
+					status.Error(codes.Internal,
+						fmt.Sprintf("Failed to check if snapshot's parent volume %s with ID %s exist, err: %s",
+							existingSnap.VolumeName, existingSnap.VolumeID, err.Error()))
+			}
+
+			// Get existing volume device info
+			parentVolFsType, err := driver.flavor.GetVolumePropertyOfPV("fsType", existingParentVolume.Name)
+			if err != nil {
+				log.Error("err: ", err.Error())
+				return nil,
+					status.Error(codes.Internal,
+						fmt.Sprintf("Failed to check if filesystme exists on the %s parent volume, err: %s",
+							existingSnap.VolumeName, err.Error()))
+			}
+
+			// Check if requested filesystem for a clone volume is same as existing snapshot
+			if filesystem != parentVolFsType {
+				return nil,
+					status.Error(codes.InvalidArgument,
+						fmt.Sprintf("Requested volume filesystem %s cannot be different than snapshot's parent volume filesystem %s", filesystem, parentVolFsType))
+			}
 			// Create a clone from another volume
 			log.Infof("About to create a new clone '%s' from snapshot %s with options %+v", name, existingSnap.ID, createOptions)
 			volume, err := storageProvider.CloneVolume(name, description, "", existingSnap.ID, size, createOptions)
--- a/pkg/driver/controller_server.go
+++ b/pkg/driver/controller_server.go
@@ -465,6 +465,13 @@ func (driver *Driver) createVolume(
 							existingSnap.VolumeName, existingSnap.VolumeID, err.Error()))
 			}
 
+			if existingParentVolume == nil {
+				return nil,
+					status.Error(codes.NotFound,
+						fmt.Sprintf("Could not find Snapshot's parent volume %s with ID %s",
+							existingSnap.VolumeName, existingSnap.VolumeID))
+
+			}
 			// Get existing parent volume fsType attribute
 			parentVolFsType, err := driver.flavor.GetVolumePropertyOfPV("fsType", existingParentVolume.Name)
 			if err != nil {
--- a/pkg/driver/controller_server.go
+++ b/pkg/driver/controller_server.go
@@ -471,7 +471,7 @@ func (driver *Driver) createVolume(
 				log.Error("err: ", err.Error())
 				return nil,
 					status.Error(codes.Internal,
-						fmt.Sprintf("Failed to check if filesystme exists on the %s parent volume, err: %s",
+						fmt.Sprintf("Failed to check if filesystem exists on the %s parent volume, err: %s",
 							existingSnap.VolumeName, err.Error()))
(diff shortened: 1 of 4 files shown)


pkg/flavor/kubernetes/flavor.go
```diff
@@ -590,3 +590,20 @@ func (flavor *Flavor) GetCredentialsFromPodSpec(volumeHandle string, podName str
 	}
 	return nil, fmt.Errorf("Pod %s/%s does not contain the volume %s", namespace, podName, volumeHandle)
 }
+
+// GetFsTypeOfPvName retrieves volume filesystem for a given CSI volname
+func (flavor *Flavor) GetFsTypeOfPvName(name string) (string, error) {
```
raunakkumar: nit: PVName

pkg/driver/controller_server.go
```diff
@@ -457,6 +455,38 @@ func (driver *Driver) createVolume(
...
+						fmt.Sprintf("Requested volume size %d cannot be lesser than snapshot's parent volume size %d", size, existingParentVolume.Size))
+			}
+
+			// Get existing volume device info
+			parentVolFsType, err := driver.flavor.GetFsTypeOfPvName(existingParentVolume.Name)
+			if err != nil {
+				log.Error("err: ", err.Error())
+				return nil,
+					status.Error(codes.Internal,
+						fmt.Sprintf("Failed to check if snapshot's parent volume %s with ID %s filesystem exist, err: %s",
```
raunakkumar: nit: change to `Failed to check if filesystem exists on the  parent volume..`

pkg/driver/controller_server.go
```diff
@@ -457,6 +455,38 @@ func (driver *Driver) createVolume(
...
+			existingParentVolume, err := storageProvider.GetVolume(existingSnap.VolumeID)
+			if err != nil {
+				log.Error("err: ", err.Error())
+				return nil,
+					status.Error(codes.Internal,
+						fmt.Sprintf("Failed to check if snapshot's parent volume %s with ID %s exist, err: %s",
+							existingSnap.VolumeName, existingSnap.VolumeID, err.Error()))
+			}
+			// The requested size is must be at least equal to the snapshot's parent volume size
+			if size < existingParentVolume.Size {
```
raunakkumar: This condition is checked in the CSP so can be removed i.e the size check

pkg/flavor/kubernetes/flavor.go
rgcostea: Let's not limit ourselves to one property.  What if we make this more generic so it can apply to any property we might want to read?  Something like:

`GetPropertyOfPv(property, pvName string)`

vidhutsingh: Sure thx!

pkg/driver/controller_server.go
vidhutsingh: ok

pkg/driver/controller_server.go
vidhutsingh: Got it !

pkg/flavor/kubernetes/flavor.go
vidhutsingh: Sorry, I didnt understand by all property of PV,  shouldn't we limit ourselves to PV's CSI property or VolumeAttributes only even if we want to make it more generic?

vidhutsingh: Made it generic for volumeAttribute properties.

pkg/driver/controller_server.go
```diff
@@ -457,6 +455,32 @@ func (driver *Driver) createVolume(
...
+							existingSnap.VolumeName, existingSnap.VolumeID, err.Error()))
+			}
+
+			// Get existing volume device info
+			parentVolFsType, err := driver.flavor.GetVolumePropertyOfPV("fsType", existingParentVolume.Name)
+			if err != nil {
+				log.Error("err: ", err.Error())
+				return nil,
+					status.Error(codes.Internal,
+						fmt.Sprintf("Failed to check if filesystme exists on the %s parent volume, err: %s",
```
rgcostea: filesystem

vidhutsingh: done.thx

pkg/driver/controller_server.go
```diff
@@ -457,6 +455,32 @@ func (driver *Driver) createVolume(
...
+			if err != nil {
+				log.Error("err: ", err.Error())
+				return nil,
+					status.Error(codes.Internal,
+						fmt.Sprintf("Failed to check if snapshot's parent volume %s with ID %s exist, err: %s",
+							existingSnap.VolumeName, existingSnap.VolumeID, err.Error()))
+			}
+
+			// Get existing volume device info
+			parentVolFsType, err := driver.flavor.GetVolumePropertyOfPV("fsType", existingParentVolume.Name)
```
shivamerla: Please add a nil check for existingParentVolume here. When not found, this will be nil along with err=nil.

pkg/flavor/kubernetes/flavor.go
```diff
@@ -590,3 +590,21 @@ func (flavor *Flavor) GetCredentialsFromPodSpec(volumeHandle string, podName str
...
+	if err != nil {
+		log.Errorf("Error retrieving the attribtue %s of the pv %s, err: %v", propertyName, pvName, err.Error())
+		return "", err
+	}
+	volAttr := pv.Spec.CSI.VolumeAttributes
+
+	if propertyVal, found := volAttr[propertyName]; found {
+		return propertyVal, nil
+	}
+	return "", fmt.Errorf("unable to retrieve attribute %s of the pv %s : not found", propertyName, pvName)
```
shivamerla: its better to return "", nil as its really not an error if the option is not found. Caller has to check both values for nil before using.

pkg/driver/controller_server.go
vidhutsingh: existingParentVolume nil check is handled above . Not sure if you meant that.

vidhutsingh: existingParentVolume nil check is handled above . Not sure if you meant that.

pkg/flavor/kubernetes/flavor.go
vidhutsingh: Accepted.

pkg/driver/controller_server.go
shivamerla: Where is it handled? i don't see this after line 459:

pkg/driver/controller_server.go
```diff
@@ -457,6 +455,32 @@ func (driver *Driver) createVolume(
...
+			// Get the parent volume from the snapshot
+			existingParentVolume, err := storageProvider.GetVolume(existingSnap.VolumeID)
+			if err != nil {
+				log.Error("err: ", err.Error())
+				return nil,
+					status.Error(codes.Internal,
+						fmt.Sprintf("Failed to check if snapshot's parent volume %s with ID %s exist, err: %s",
+							existingSnap.VolumeName, existingSnap.VolumeID, err.Error()))
+			}
+
```
suneeth51: As @shivamerla pointed, 'GetVolume' is expected to return both 'existingParentVolume' and 'err' as nil when volume is not found. You have already handled err != nil case, but still need to handle the other case when 'existingParentVolume' is nil.

Please add the below check: 
if existingParentVolume == nil {
	return nil, status.Error(codes.NotFound, "Could not find parent volume with ID "+volID)
}

vidhutsingh: Ok, thanks. I will handle it although this is not my code change and it has probably come from upstream after merge. 
Here is the relevant PR . 
https://github.com/hpe-storage/csi-driver/pull/108/commits/7c9aebc84d892a01bce77be363ee24a9ba0ad372

pkg/driver/controller_server.go
vidhutsingh: Please check the updated code. thx

shivamerla: looks good. Thanks.

vidhutsingh:
Thank you Shiva for merging the changes !

PR #112, opened 2020-03-02
Fix sanity test failure 

vidhutsingh:
Problem:
sanity test works against fake csp driver and it doesnt have fsType for parent volume. 

Fix:
Added additional check fsType != "" to avoid sanity test failure.


shivamerla:
@vidhutsingh can you rebase and send only the change please? Also, did you try make test to verify no other failures?

vidhutsingh:
I have rebased it.

vidhutsingh:
> @vidhutsingh can you rebase and send only the change please? Also, did you try make test to verify no other failures?

Please check the latest, I rebased it even last time I ran "make all" which eventually runs "test" as well.

PR #113, opened 2020-03-02
csi: Fix volume expansion for raw block volume

shivamerla:
* Problem:
  * With raw block volume, volume path is /dev/dm-x and we are attempting to
  * find staging device file with this path and return error.
* Implementation:
  * Get staging device info only for mount access mode, and directly pass the
  * volume path for node expand without staging device info.
* Testing: CSI certification tests pass with this fix.
* Review: gcostea, rkumar
* Bug: https://nimblejira.nimblestorage.com/browse/NLT-

### DIFF
--- a/pkg/driver/node_server.go
+++ b/pkg/driver/node_server.go
@@ -1528,7 +1528,7 @@ func (driver *Driver) NodeExpandVolume(ctx context.Context, request *csi.NodeExp
 		// figure out if volumePath is actually a staging path
 		stagedDevice, err := readStagedDeviceInfo(request.GetVolumePath())
 		if err != nil {
-			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Cannot get staging device info from volume path. %s", err.Error()))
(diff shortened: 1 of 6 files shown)


pkg/driver/node_server.go
```diff
@@ -1502,36 +1502,52 @@ func (driver *Driver) NodeExpandVolume(ctx context.Context, request *csi.NodeExp
...
+		targetPath = request.GetVolumePath()
+		// Expand device to underlying volume size
+		log.Infof("About to expand device %s with access type block to underlying volume size",
+			request.VolumePath)
+		err = driver.chapiDriver.ExpandDevice(targetPath, model.BlockType)
+	} else {
+		// figure out if volumePath is actually a staging path
+		stagedDevice, err := readStagedDeviceInfo(request.GetVolumePath())
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Cannot get staging device info from volume path. %s", err.Error()))
```
rgcostea: nit: drop the . after path

PR #132, opened 2020-03-14
Remove data wrapper

rgcostea:
Verified PVC and Pod creation


rgcostea:
Yes. Only vendor update

PR #134, opened 2020-04-02
Serialize NodePublish()/NodeUnpublish() workflow for ephemeral volumes and ephemeral vol rename

suneeth51:
1)  CON-852: Rename ephemeral volume by appending PodName.
     - Format:  **ephemeral-<PODNAME>-<VolumeHandle>**
     - The PodName infix will be truncated to 32 characters if it's larger.

2) CON-839:Serialized the NodePublish()/NodeUnpublish workflows using Mutex locks for ephemeral volumes.

### DIFF
--- a/pkg/driver/node_server.go
+++ b/pkg/driver/node_server.go
@@ -28,8 +28,10 @@ import (
 )
 
 var (
-	stageLock   sync.Mutex
-	unstageLock sync.Mutex
+	stageLock              sync.Mutex
+	unstageLock            sync.Mutex
+	ephemeralPublishLock   sync.Mutex
+	ephemeralUnpublishLock sync.Mutex
 )
 
 // Helper utility to construct default mountpoint path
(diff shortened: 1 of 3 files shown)


pkg/driver/node_server.go
```diff
@@ -28,8 +28,10 @@ import (
 )
 
 var (
-	stageLock   sync.Mutex
-	unstageLock sync.Mutex
+	stageLock        sync.Mutex
+	unstageLock      sync.Mutex
+	ephPublishLock   sync.Mutex
```
raunakkumar, line 33: nit: you can name it ephemeral :)

pkg/driver/node_server.go
```diff
@@ -926,13 +945,11 @@ func (driver *Driver) nodePublishVolume(
 // retry the rollback on failure
 func (driver *Driver) retryRollbackEphemeralVolume(
 	volumeHandle string,
+	volumeName string,
 	secrets map[string]string,
 	stagingTargetPath string,
 	targetPath string) error {
 
-	// Get the volume if exists
```
raunakkumar, line 933: why did we remove this? Any issues?

suneeth51: > why did we remove this? Any issues?

As we are passing the `volumeName` from the caller, we don't need to construct the name again here.

pkg/driver/node_server.go
```diff
@@ -28,8 +28,10 @@ import (
 )
 
 var (
-	stageLock   sync.Mutex
-	unstageLock sync.Mutex
+	stageLock        sync.Mutex
+	unstageLock      sync.Mutex
+	ephPublishLock   sync.Mutex
+	ephUnpublishLock sync.Mutex
```
rgcostea, line 34: Agreed.  Spell out ephemeral.

PR #137, opened 2020-04-03
ephemeral volume, make accessProtocol mandatory

raunakkumar:
* Problem:
 generic error Failed to get storage provider from secrets, No secrets have been provided
* Implementation:
 prevent ephemeral volume creation if accessProtocol is not specified
* Testing:
* Bug: https://nimblejira.nimblestorage.com/browse/CON-862

### DIFF
--- a/examples/kubernetes/pod-inline-with-vol-secrets.yaml
+++ b/examples/kubernetes/pod-inline-with-vol-secrets.yaml
@@ -20,4 +20,5 @@ spec:
          csi.storage.k8s.io/ephemeral: "true"
          inline-volume-secret-name: nimble-secret
          inline-volume-secret-namespace: kube-system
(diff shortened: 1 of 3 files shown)


drajen:
May I suggest we have the default fallback set to "iscsi"? Even contemplating if annotating the `Secret` to have a `hpe.com/defaultAccessProtocol` to trickle down on both `StorageClass` and the inline volumes to put less care to the end user of the system.

raunakkumar:
> May I suggest we have the default fallback set to "iscsi"? Even contemplating if annotating the `Secret` to have a `hpe.com/defaultAccessProtocol` to trickle down on both `StorageClass` and the inline volumes to put less care to the end user of the system.

we already had the default to `iscsi`, the problem was seen on `fc` environments without any meaningful errors

PR #138, opened 2020-04-05
Report volume limits per node with GetNodeInfo configured with deployment

vidhutsingh:
Feature: CSI Volume Limit 
Requirement : 
To support volume limits in a CSI driver, the plugin must fill in max_volumes_per_node in NodeGetInfoResponse. 

Implementations:
k8s cluster node supports max_volume_per_node though, this property should be advertised by Node driver during NodeGetInfoResponse. The CSI driver implementation will use node env 
variable MAX_VOLUME_PER_NODE if this is set by the user during deployment. If user doesn't set , default value 100 Volume Limit will be used by the CSI driver. 

Example : 
  - name: hpe-csi-driver
          image: dcs-docker.sjcartifactory.eng.nimblestorage.com/hpestorage/csi-driver:vsingh_priv
          args :
            - "--endpoint=$(CSI_ENDPOINT)"
            - "--node-service"
            - "--flavor=kubernetes"
          env:
            - name: CSI_ENDPOINT
              value: unix:///csi/csi.sock
            - name: LOG_LEVEL
              value: trace
            - name: MAX_VOLUMES_PER_NODE
              value: "20"

### DIFF
--- a/pkg/driver/node_server.go
+++ b/pkg/driver/node_server.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -33,6 +34,10 @@ var (
 	ephemeralPublishLock   sync.Mutex
 	ephemeralUnpublishLock sync.Mutex
 )
+
+// Maximum default number of volumes that controller can publish to the node.
+const defaultMaxVolPerNode = 100
+const maxVolumePerNodeKey = "MAX_VOLUMES_PER_NODE"
 
 // Helper utility to construct default mountpoint path
 func getDefaultMountPoint(id string) string {
@@ -1608,10 +1613,39 @@ func (driver *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	// Get max volume per node from environment variable
+	nodeMaxVolumesLimit := driver.nodeGetIntEnv(maxVolumePerNodeKey)
+
+	// Maximum number of volumes that controller can publish to the node.
+  	// If value is not set or zero CO SHALL decide how many volumes of
+  	// this type can be published by the controller to the node. The
+  	// plugin MUST NOT set negative values here.
+ 	// This field is OPTIONAL.
+	if nodeMaxVolumesLimit <= 0 {
+		nodeMaxVolumesLimit = defaultMaxVolPerNode
+	}
 
 	return &csi.NodeGetInfoResponse{
-		NodeId: nodeID,
+		NodeId:            nodeID,
+		MaxVolumesPerNode: nodeMaxVolumesLimit,
 	}, nil
+}
+
+// Get cluster node environment variable, convert it to int64
+func (driver *Driver) nodeGetIntEnv(envKey string) int64 {
+	// Read max_volume_per_node from env
+	envVal := os.Getenv(envKey)
+	// Check if env variable is set by user
+	if envVal == "" {
+		return 0
+	}
+	val, err := strconv.ParseInt(envVal, 10, 64)
+	if err != nil {
+		log.Warnf("Failed to read cluster node %s setting. Error: %s", envKey, err.Error())
+		return 0
+
+	}
+	return val
 }
 
 // nolint: gocyclo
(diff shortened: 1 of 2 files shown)


pkg/driver/node_server.go
```diff
@@ -34,6 +35,10 @@ var (
 	ephemeralUnpublishLock sync.Mutex
 )
 
+// Maximum default number of volumes that controller can publish to the node.
+const defaultMaxVolPerNode = 100
+const maxVolumePerNodeKey = "MAX_VOLUMES_PER_NODE"
+
```
rgcostea, line 41: We have a constants.go file where we keep these.

pkg/driver/node_server.go
```diff
@@ -1608,12 +1613,41 @@ func (driver *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
...
+	envVal := os.Getenv(envKey)
+	// Check if env variable is set by user
+	if envVal == "" {
+		return 0
+	}
+	val, err := strconv.ParseInt(envVal, 10, 64)
+	if err != nil {
+		log.Warnf("Failed to read cluster node %s setting. Error: %s", envKey, err.Error())
+		return 0
+
```
rgcostea, line 1646: nit: extra line

pkg/driver/node_server.go
```diff
@@ -34,6 +35,10 @@ var (
 	ephemeralUnpublishLock sync.Mutex
 )
 
+// Maximum default number of volumes that controller can publish to the node.
+const defaultMaxVolPerNode = 100
+const maxVolumePerNodeKey = "MAX_VOLUMES_PER_NODE"
```
shivamerla, line 40: nit: maxVolumesPerNodeKey

shivamerla:
@vidhutsingh as part of your testing, also verify if k8s is actually honoring these limits. You can set limits to be very low(say 1) and try to run pods > nodes. Remaining should stay in pending state. Later you can dynamically change limits(re-deploy pods) to see if they can be scheduled again. Also other test is run few pods per node initially with default limit, and later change the limit to 1 dynamically. Existing pods should continue to run and on next drain/pod restart etc, they might enter pending state. We are not sure of these behaviors, so better to document these for our QA.

pkg/driver/node_server.go
vidhutsingh: Sorry I didn't notice. I will use it.

pkg/driver/node_server.go
vidhutsingh: Sure.

vidhutsingh:
> @vidhutsingh as part of your testing, also verify if k8s is actually honoring these limits. You can set limits to be very low(say 1) and try to run pods > nodes. Remaining should stay in pending state. Later you can dynamically change limits(re-deploy pods) to see if they can be scheduled again. Also other test is run few pods per node initially with default limit, and later change the limit to 1 dynamically. Existing pods should continue to run and on next drain/pod restart etc, they might enter pending state. We are not sure of these behaviors, so better to document these for our QA.

Thanks Shiva,  I will provide the expected behavior for all the scenarios.

pkg/driver/node_server.go
vidhutsingh: done.

vidhutsingh:
When Volume Limit is hit on Node , K8 scheduler throw following warning message 
"Warning  FailedScheduling  <unknown>  default-scheduler  0/3 nodes are available: 1 node(s) had taints that the pod didn't tolerate, 2 node(s) exceed max volume count."

shivamerla:
> When Volume Limit is hit on Node , K8 scheduler throw following warning message
> "Warning FailedScheduling default-scheduler 0/3 nodes are available: 1 node(s) had taints that the pod didn't tolerate, 2 node(s) exceed max volume count."

Thanks Vidhut. Also, when you change the limits and redeploy CSI driver, they will get scheduled automatically? Can you add these details to the confluence page with brief details about this feature.

vidhutsingh:
> > When Volume Limit is hit on Node , K8 scheduler throw following warning message
> > "Warning FailedScheduling default-scheduler 0/3 nodes are available: 1 node(s) had taints that the pod didn't tolerate, 2 node(s) exceed max volume count."
> 
> Thanks Vidhut. Also, when you change the limits and redeploy CSI driver, they will get scheduled automatically? Can you add these details to the confluence page with brief details about this feature.

Yes, Shiva. when limit is increased, the POD got scheduled automatically. 
 Let me captured this in confluence and share it across.

PR #141, opened 2020-04-09
Update NFS ganesha image to 2.8 and usinng UBI image

shivamerla:


pkg/flavor/kubernetes/nfs.go
```diff
@@ -23,7 +23,7 @@ import (
 const (
 	nfsPrefix    = "hpe-nfs-"
 	nfsNamespace = "hpe-nfs"
-	nfsImage     = "nimblestorage/nfs-ganesha:2.5.0"
+	nfsImage     = "hpestorage/nfs-provisioner:2.8.3-4"
```
rgcostea, line 26: Is this an image we built?

PR #144, opened 2020-04-15
csi-driver: cleanup code for cloneOfPVC

shivamerla:
* Problem:
  * Currently cloneOfPVC assumes PV as an input and confusing
  * with cloneOf parameter.
* Implementation:
  * Deprecate and cleanup of cloneOfPVC option in favor of cloneOf
  * which refers to source PV to clone from.
* Testing: tested with cloneOf to make sure for no regressions.
* Review: gcostea, rkumar
* Bug: https://nimblejira.nimblestorage.com/browse/NLT-

### DIFF
--- a/pkg/flavor/kubernetes/flavor.go
+++ b/pkg/flavor/kubernetes/flavor.go
@@ -411,43 +411,11 @@ func (flavor *Flavor) getClaimOverrideOptions(claim *v1.PersistentVolumeClaim, o
 		}
 	}
 
-	// check to see if there is cloneOfPVC annotation present
-	volumeName, err := flavor.getPVFromPVCAnnotation(claim, provisioner)
-	if err != nil {
-		return nil, err
-	}
-
-	// update the options map with the pv name if there is one
-	if volumeName != "" {
-		optionsMap[cloneOf] = volumeName
-	}
-
 	// make sure cloneOfPVC is removed from the options (all cases)
 	if _, found := optionsMap[cloneOfPVC]; found {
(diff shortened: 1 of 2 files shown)


pkg/flavor/kubernetes/flavor.go
```diff
@@ -411,45 +411,13 @@ func (flavor *Flavor) getClaimOverrideOptions(claim *v1.PersistentVolumeClaim, o
...
-
-	// update the options map with the pv name if there is one
-	if volumeName != "" {
-		optionsMap[cloneOf] = volumeName
-	}
-
 	// make sure cloneOfPVC is removed from the options (all cases)
 	if _, found := optionsMap[cloneOfPVC]; found {
 		delete(optionsMap, cloneOfPVC)
 	}
```
rgcostea, line 417: Why don't we drop this too?  Along with the constant?

shivamerla: I just kept it in case if anyone trying to use this option  from flexvolume references. CSP will ignore it in that case?

rgcostea: But this is the CSI code base.  Support for this is implemented via DataSource.

shivamerla: sure, i just tested by passing cloneOfPVC without removing from option map. will cleanup this too.

shivamerla: done..

PR #145, opened 2020-04-16
Implemented generic os event driven mechanism to update HPENodeInfo object if there is any change in object attributes detected .

vidhutsingh:
**Issue:**
CON-882 (Continuously monitor worker node network configuration)

**Implementations:**
Implemented a generic polling mechanism that accepts anonymous function and polling interval as arguments. 
It has two functions 
1.obj:=InitTask(...)

This is basically used to initialize the task object which registers the anonymous function that needs to be polled along with poll frequency. Also, it invokes a separate thread to monitor the OS signals e.g SIGTERM, SIGHUP for the graceful exit of the polling task. 

2.obj.start()
This is basically to kick starts the polling in a loop unit os signal is received. It periodically polls the registered anonymous fn. This should be invoked as a separate goroutine.

Created an anonymous wrapper function around getNodeInfo() that is passed to InitTask() with default 30 Minutes polling frequency. 

**Testing:**
I have tested the code by manually deleting the cluster worker node's one of the data network interfaces and have validated that the network changes are dynamically picked by Node Server and also HPENodeInfo is updated.


pkg/driver/csi_poll.go
```diff
@@ -0,0 +1,91 @@
...
+		syscall.SIGTERM,
+		syscall.SIGHUP,
+		syscall.SIGTERM,
+		syscall.SIGKILL,
+	)
+	// Create a thread to monitor the os signals.
+	go func() {
+		select {
+		case sig := <-sigc:
+			log.Info("Received %s os signal. Exiting...\n", sig)
```
raunakkumar, line 56: nit: Infof

vidhutsingh: Done. thx

pkg/driver/node_server.go
```diff
@@ -1621,6 +1622,20 @@ func (driver *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
...
+	// Create a anonymous wrapper function over nodeGetInfo. This function is called
+	// periodically.
+	getNodeInfoFunc := func() {
+		nodeID, err = driver.nodeGetInfo()
+		if err != nil {
+			log.Errorf("Failed to update %s nodeInfo. Error: %s", nodeID, err.Error())
+		}
+		return
+	}
+	// Initialize polling with anonymous wrapper function.
```
shivamerla, line 1634: I think GetNodeInfo might not be the right place to add this task, as there can be possibility for kubelet to call this function multiple times adding duplicate tasks. Can we add this as part of NodeDriver init itself?

vidhutsingh: Good catch Shiva. thanks. 
CO may call this GetNodeInfo rpc more than once although I have not noticed during testing. As mostly CO invokes NodeGetCapabilities rpc multiple times. I will make the code changes .

pkg/driver/constants.go
```diff
@@ -87,4 +87,6 @@ const (
 	// Maximum default number of volumes that controller can publish to the node.
 	defaultMaxVolPerNode = 100
 	maxVolumesPerNodeKey = "MAX_VOLUMES_PER_NODE"
+	// GetNodeInfo polling frequency
+	PollingFrequencyInMin = 30
```
drajen, line 91: Where are we getting the 30 minute interval from? How expensive is this? Can we poll every minute?

What I'm getting is that if there's a network change, that change will immediately be tested by an end-user and they might start digging themselves a hole because no one is going to sit on their hands for 30 minutes nor read the documentation to understand this behaviour. If we poll every minute, a few retries down the road and you're up and running.

vidhutsingh: Polling every minute may put extra load on the cluster node if there is no frequent network change. 30 min is default polling delay as you see in the code.  I agree that any change to network configuration will be picked only after 30 min which may not be acceptable to the user.
We could add a configurable env variable for this. I will let @rgcostea  comment on this. thanks.

rgcostea: So I came across fsnotify last night: https://godoc.org/github.com/fsnotify/fsnotify

Could we use this to set a watcher on /etc/sysconfig/network and/or /etc/sysconfig/network-scripts and/or any other files we think might change as users update their networks?  This way, it is notification driven rather polled.

shivamerla: It is a better approach, but i think more challenging. We would need watch list of directories based on distro(/etc/network, /etc/sysconfig/network, /etc/iscsi/) for changes. The challenge here is some tool on host should not be generating spurious updates very often to any files under these without actual network change.

rgcostea: Even if spurious updates come along, that doesn't mean we have to update the node info on each one.  We could be selective.  For example, we could read the files of interest and compare against our current node to determine if we need to perform an update.  Furthermore, we could coalesce multiple updates into one operation on our end.

shivamerla: sure, sounds good.

vidhutsingh: The fsnotify looks like a promising package that is os event-driven and useful for monitoring any file changes on host os that is different then the original requirement (captured in the Bug) to poll for networking changes and update the HPENodInfo.  We should evaluate the following points before we take this route : 
1. HPENodeInfo doesn't not only contain networking , it also contains  initiators name, domain name and chap info, is this requirement to monitor all of them or only networking ? 
2.  If there is any change in default configuration file path on host , this will not work.  I don't think there would be lot of such users.

vidhutsingh: I'm working on this change.

rgcostea: Yes, we should monitor for every property that can change including IQN and WWPN details.

Item 2 is a corner case as you say.  So in those cases, we just fall back to our current approach of restarting the node driver.

vidhutsingh: Thanks @rgcostea . Done.

pkg/driver/csi_watch.go
```diff
@@ -0,0 +1,116 @@
...
+	watchStop chan struct{}
+	// fsnotify watcher.
+	watchList *notify.Watcher
+	// Anonymous function.
+	watchRun func()
+	// Wait
+	wg sync.WaitGroup
+}
+
+// InitializeWatcher is used to initialized CSIWatch with anonymous function and new watcher.
```
rgcostea, line 27: typo: used to initialize CSIWatch

pkg/driver/csi_watch.go
```diff
@@ -0,0 +1,116 @@
+// Copyright 2019 Hewlett Packard Enterprise Development LP
```
rgcostea, line 1: 2020

pkg/driver/csi_watch.go
```diff
@@ -0,0 +1,116 @@
...
+			// Call stopWatcher() for gracefull exit of watcher.
+			task.stopWatcher()
+			task.wg.Wait()
+		}
+	}()
+
+	return task, nil
+}
+
+//AddWatchList list of files /and direcror to watch
```
rgcostea, line 69: files and directories

pkg/driver/csi_watch.go
```diff
@@ -0,0 +1,116 @@
...
+			log.Warnf("Failed to add [%s] file to watch list, err :", fPath, err.Error())
+		} else {
+			log.Tracef("Successfully added [%s] file to watch list", fPath)
+		}
+
+	}
+	return nil
+}
+
+// Start polling until os sig intterupt. This will run anonymous fn forever.
```
rgcostea, line 89: typo: interrupt

pkg/driver/csi_watch.go
```diff
@@ -0,0 +1,116 @@
...
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+
+	notify "github.com/fsnotify/fsnotify"
+	log "github.com/hpe-storage/common-host-libs/logger"
+)
+
+type CSIWatch struct {
```
rgcostea, line 16: This seems to be a generic watcher unrelated to CSI.  Can we move this to common-host-libs and rename it to FileWatcher?

pkg/driver/node_server.go
```diff
@@ -34,6 +34,7 @@ var (
 	ephemeralPublishLock   sync.Mutex
 	ephemeralUnpublishLock sync.Mutex
 )
+var isWatcherEnable bool = false
```
rgcostea, line 37: nit: isWatcherEnabled

pkg/driver/node_server.go
```diff
@@ -1621,6 +1623,28 @@ func (driver *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
...
+			return
+		}
+		// Register anonymous wrapper function(watcher).
+		csiWatch, _ := driver.InitializeWatcher(getNodeInfoFunc)
+		// Add list of files /and directories to watch. The list contains
+		// iSCSI , FC and CHAP Info and Networking config directories
+		list := []string{"/etc/sysconfig/network-scripts/", "/etc/sysconfig/network/", "/etc/iscsi/", "/sys/class/fc_host/"}
+		csiWatch.AddWatchList(list)
+		// Start event the watcher in a separate thread.
+		go csiWatch.StartWatcher()
```
rgcostea, line 1645: Per @shivamerla , how are we handling spurious updates?  We don't want the getNodeInfoFunc to be called on every update to every file in these directories.

pkg/driver/node_server.go
```diff
@@ -1621,6 +1623,28 @@ func (driver *Driver) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
...
+	if !isWatcherEnable {
+		// Create a anonymous wrapper function over nodeGetInfo. This is os event driven
+		// fn execution.
+		getNodeInfoFunc := func() {
+			nodeID, err := driver.nodeGetInfo()
+			if err != nil {
+				log.Errorf("Failed to update %s nodeInfo. Error: %s", nodeID, err.Error())
+			}
+			return
+		}
```
rgcostea, line 1637: I feel this is overkill.  The watcher can determine which files have changed and therefore limit how much of the node info needs to be updated.  For example, why update the entire node when only the iqn has changed.  Likewise, why update the entire node when only the network has changed?

This might be difficult to achieve so at least minimizing the frequency with which this function is called would probably be enough.  So can you add a logger to this function so we can start gauging that?

pkg/driver/constants.go
shivamerla: @vidhutsingh please add directories /etc/network, /etc/sysconfig-networks, /etc/network-scripts and file /etc/iscsi/InitiatorName and remove /etc/iscsi from monitoring. /etc/iscsi will have iscsi DB on Ubuntu which gets updated on every volume attach/detach, but we just need to capture IQN change. We also discussed that we need rate limiting on this(i.e process only one change request every 5 minutes etc). For FC, i don't think we need monitoring as new adapter(WWN) involves host reboot anyway.

vidhutsingh: Thanks shiva. sure I will make watcher list changes. 
Regarding, If we rate limit to 5 min than how different is this implementation than earlier polling implementation?

pkg/driver/csi_watch.go
vidhutsingh: Done

pkg/driver/csi_watch.go
vidhutsingh: Done.

pkg/driver/csi_watch.go
vidhutsingh: Done.

pkg/driver/csi_watch.go
vidhutsingh: Yes, this is a generic implementation and doesn't have any CSI specific implementation. This will be useful if at any other place we want to have an event-driven logic. I will move this to common-host-libs. thanks

pkg/driver/node_server.go
vidhutsingh: Done.

pkg/driver/node_server.go
vidhutsingh: We aren't updating the entire HPENodeInfo if you check LoadNodeInfo() in flavor.go, only specific changes are updated. So if there is only network change we do network update only. Please check and let me know.

pkg/driver/node_server.go
vidhutsingh: I am capturing all the events trigger on WatchList, However, we could go even more granular in our implementation like only check for CREATE event in network directory. Please elaborate on spurious updates so that we can safeguard against it.

pkg/driver/csi_watch.go
vidhutsingh: Do you want me to create a separate package FileWatcher inside common-host-libs ? Since none of the existing package names suits watcher implementation.

rgcostea: I think it fits in the util package.

vidhutsingh: ok thanks.

pkg/driver/csi_watch.go
vidhutsingh: done.thx

pkg/driver/node_server.go
rgcostea: Right.  That's LoadNodeInfo.  But that is preceded by nodeGetInfo which calls driver.chapiDriver.GetHostNameAndDomain, driver.chapiDriver.GetHostInitiators, and driver.chapiDriver.GetHostNetworks.  It seems we should only invoke CHAPI when necessary.

pkg/driver/constants.go
rgcostea: It's different from polling because we would still only update when there is a change to these directories.  If no changes are made from 10pm to 8am the next morning, we wouldn't do anything.  The same cannot be said for polling.

pkg/driver/node_server.go
rgcostea: This is along the lines of rate limiting.  We don't want to call nodeGetInfo on every update.  Only on all updates within a time interval.

pkg/driver/node_server.go
vidhutsingh: Yes, I think we should also consider that this is event-driven and we aren't expecting a regular update to Network, Initiators or Domain, etc. So, even if we call nodeGetInfo() that should be okay since it will not have a major performance impact. Thought?

pkg/driver/constants.go
vidhutsingh: Okay, thanks for the clarification. so basically we should go to sleep mode for 5 min after invoking t.watchRun() . 
So after every event notification we will run the nodeGetInfo() and go to sleep for 5 min and if there is any event in between that will be ignored. That will rate limit.

pkg/driver/node_server.go
vidhutsingh: Got it. thanks.

pkg/driver/constants.go
vidhutsingh: The corner case still exist, what if there is a genuine network update within 5 min window? that will be ignored. shall we stick to 1 min instead of 5 min? Thought?

pkg/driver/csi_watch.go
vidhutsingh: https://github.com/hpe-storage/common-host-libs/pull/112

pkg/driver/constants.go
vidhutsingh: @shiva, we need /etc/iscsi/iscsid.conf file for CHAP update.

pkg/driver/node_server.go
rgcostea: Ya, that's fair.  I don't think this is a hard requirement, just ideal.  I think if we limit the overall frequency at which we update (only when necessary), this should be fine.

pkg/driver/constants.go
rgcostea: Where did 5 minutes come from?  :)

I think 1 minute would be better.

vidhutsingh: It's been suggested above. thanks for the confirmation.

shiva:
Hi,

I'm not sure how to get these email to stop. Can you please remove me from
your mailing list?

-Shiva

vidhutsingh:
> Hi, I'm not sure how to get these email to stop. Can you please remove me from your mailing list?
> […](#)
> -Shiva


I guess you need to Unsubscribe notification. That should stop. Since I haven't made any user-specific configuration it was the default.

shivamerla:
> > Hi, I'm not sure how to get these email to stop. Can you please remove me from your mailing list?
> > […](#)
> > -Shiva
> 
> I guess you need to Unsubscribe notification. That should stop. Since I haven't made any user-specific configuration it was the default.

@shiva  Looks like folks have been tagging you instead of me by mistake. My GitHub ID is @shivamerla. I am not sure how i can control it. Can you unsubscribe from notifications on this repo?

shiva:
Will try. Thanks for your responses.

-Shiva

PR #152, opened 2020-04-28
csi-driver: use node name from the podSpec to create hpenodeinfo objects

shivamerla:
* Problem:
  * Currently CSI node driver is creating hpenodeinfos object with name of the
  * underlying host which might not follow k8s object naming conventions. This will
  * fail if hostname has uppercase letters. But kubelet seems to convert them to lowercase to join
  * them to cluster.
* Implementation:
  * Use NODE_NAME env populated with pod fields to consider as object name for hpenodeinfos type.
  * For upgrades, even if underlying hostname differs with spec.nodeName, we will continue to use
  * already created one. For new ones, we use spec.nodeName.
* Testing: tested with deploying node plugin with changes.
* Review: gcostea, rkumar

### DIFF
--- a/pkg/flavor/kubernetes/flavor.go
+++ b/pkg/flavor/kubernetes/flavor.go
@@ -130,6 +130,11 @@ func (flavor *Flavor) LoadNodeInfo(node *model.Node) (string, error) {
 	log.Tracef(">>>>> LoadNodeInfo called with node %v", node)
 	defer log.Trace("<<<<< LoadNodeInfo")
 
+	// overwrite actual host name with node name from k8s to be compliant
+	if name := os.Getenv("NODE_NAME"); name != "" {
+		node.Name = name
+	}
+
 	nodeInfo, err := flavor.getNodeInfoByUUID(node.UUID)
 	if err != nil {
 		log.Errorf("Error obtaining node info by uuid %s- %s\n", node.UUID, err.Error())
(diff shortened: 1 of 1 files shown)


pkg/flavor/kubernetes/flavor.go
```diff
@@ -130,6 +130,11 @@ func (flavor *Flavor) LoadNodeInfo(node *model.Node) (string, error) {
 	log.Tracef(">>>>> LoadNodeInfo called with node %v", node)
 	defer log.Trace("<<<<< LoadNodeInfo")
 
+	// overwrite actual host name with node name from k8s to be compliant
+	if name := os.Getenv("NODE_NAME"); name != "" {
```
raunakkumar, line 134: would need to update the yaml/helm file to set this env, right ?

shivamerla: yes..

raunakkumar: cool. if for some reason it's not set should we default to existing hostname or that is already taken care of?

shivamerla: yes Raunak, if its set to empty, we will default to original behavior.

shivamerla:
> Looks good. I had assigned the bug to Vidhut but looks like you got to it too fast :)

Oh, i didn't realize it got assigned to Vidhut..:)

PR #157, opened 2020-05-04
nfs: fix nfs to have stable fs handle on pod failover

shivamerla:
* Problem:
  * 1. nfs pod was not able to get UUID of xfs filesystem and thus using device major:minor numbers
  * for file handles. Thus on pod failover major:minor number change causes fsHandle to go stale for clients
  * 2. nfs pods needs to be marked as critical automatically based on kubelet version of 1.17 or above.
  * 3. ganesha.conf needs to be configurable via configmap
* Implementation:
  * 1. Running nfs pod as privileged container, so that it can access XFS UUID through blkid. This causes
  * nfs-ganesha to use UUID for stable fsHandle across nodes.
  * 2. Add function to fetch current node kubelet version to add pod priority automatically from 1.17 onwards.
  * 3. Mount ganesha.conf as configmap volume instead of static file within image.
* Testing: tested failover/failback of NFS pods/application pods, configmap and pod priority setup.
* Review: gcostea, rkumar

### DIFF
--- a/pkg/flavor/kubernetes/nfs.go
+++ b/pkg/flavor/kubernetes/nfs.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	log "github.com/hpe-storage/common-host-libs/logger"
+	"golang.org/x/mod/semver"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	apps_v1 "k8s.io/api/apps/v1"
@@ -33,7 +34,6 @@ const (
 	defaultExportPath          = "/export"
 	nfsResourceLimitsCPUKey    = "nfsResourceLimitsCpuM"
 	nfsResourceLimitsMemoryKey = "nfsResourceLimitsMemoryMi"
-	nfsPodPriorityCriticalKey  = "nfsPodPriorityCritical"
 	nfsMountOptionsKey         = "nfsMountOptions"
 	nfsNodeSelectorKey         = "csi.hpe.com/hpe-nfs"
 	nfsNodeSelectorValue       = "true"
@@ -41,6 +41,8 @@ const (
 	nfsNamespaceKey            = "nfsNamespace"
 	nfsProvisionerImageKey     = "nfsProvisionerImage"
 	pvcKind                    = "PersistentVolumeClaim"
+	ganeshaConfigFile          = "ganesha.conf"
+	ganeshaConfigMap           = "hpe-nfs-config"
 )
 
 // NFSSpec for creating NFS resources
@@ -48,7 +50,6 @@ type NFSSpec struct {
 	volumeClaim          string
 	resourceRequirements *core_v1.ResourceRequirements
 	nodeSelector         map[string]string
-	podPriorityCritical  bool
 	image                string
 }
 
@@ -346,11 +347,6 @@ func (flavor *Flavor) GetNFSSpec(scParams map[string]string) (*NFSSpec, error) {
 		nfsSpec.nodeSelector = map[string]string{nfsNodeSelectorKey: nfsNodeSelectorValue}
 	}
 
-	// set pod scheduling priority if specified by user
-	if val, ok := scParams[nfsPodPriorityCriticalKey]; ok && val == "true" {
-		nfsSpec.podPriorityCritical = true
-	}
-
 	// use nfs provisioner image specified in storage class
 	nfsSpec.image = defaultNFSImage
 	if image, ok := scParams[nfsProvisionerImageKey]; ok {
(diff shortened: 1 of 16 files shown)


pkg/flavor/kubernetes/nfs.go
```diff
@@ -537,6 +588,27 @@ func (flavor *Flavor) GetNFSNodes() ([]core_v1.Node, error) {
...
+	log.Tracef(">>>>> getKubeletVersion")
+	defer log.Tracef("<<<<< getKubeletVersion")
+
+	node, err := flavor.kubeClient.CoreV1().Nodes().Get(os.Getenv("NODE_NAME"), meta_v1.GetOptions{})
+	if err != nil {
+		log.Errorf("unable to get current node %s information, err %s", os.Getenv("NODE_NAME"), err.Error())
+		return "", err
+	}
+	if !semver.IsValid(node.Status.NodeInfo.KubeletVersion) {
+		return "", fmt.Errorf("invalid kubelet vesion %s obtained", node.Status.NodeInfo.KubeletVersion)
```
rgcostea, line 601: version

pkg/flavor/kubernetes/nfs.go
```diff
@@ -649,7 +740,9 @@ func (flavor *Flavor) makeNFSDeployment(name string, nfsSpec *NFSSpec, nfsNamesp
 		podSpec.Spec.NodeSelector = nfsSpec.nodeSelector
 	}
 
-	if nfsSpec.podPriorityCritical {
+	// apply pod priority(supported from k8s 1.17 onwards to run critical pods in non kube-system namespace)
+	kubeletVersion, _ := flavor.getKubeletVersion()
+	if kubeletVersion != "" && semver.Compare(kubeletVersion, "v1.17.0") >= 0 {
 		podSpec.Spec.PriorityClassName = "system-cluster-critical"
 	}
```
rgcostea, line 747: What happens if you have mixed environment (some nodes on < 1.17 and some on 1.17+)?  If the pod is scheduled on a 1.17+ node and that node dies, will this priority setting cause the pod to fail to be scheduled to the < 1.17 nodes?

pkg/flavor/kubernetes/nfs.go
```diff
@@ -145,6 +153,54 @@ func (flavor *Flavor) CreateNFSVolume(pvName string, reqVolSize int64, parameter
...
+  Pseudo = /export;
+  Access_Type = RW;
+  Squash = No_Root_Squash;
+  Transports = TCP;
+  Protocols = 4;
+  SecType = "sys";
+  FSAL {
+	  Name = VFS;
+  }
+}`
```
rgcostea, line 181: Why don't we allow a user to specify this rather than generating it ourselves?  I thought we had discussed pushing some of those configuration options onto the user.

vendor/github.com/hpe-storage/common-host-libs/storageprovider/csp/container_storage_provider.go
```diff
@@ -462,6 +462,36 @@ func (provider *ContainerStorageProvider) ExpandVolume(id string, requestBytes i
 	return response, err
 }
 
+// CreateVolume creates a volume on the CSP
+func (provider *ContainerStorageProvider) EditVolume(id string, opts map[string]interface{}) (*model.Volume, error) {
```
rgcostea, line 466: Comment doesn't match the function name.

pkg/flavor/kubernetes/nfs.go
shivamerla: yes..the challenge is it has to be in every namespace where the pod is being deployed..so might get confusing. This way, we still automatically create a configmap if one doesn't exist, and user can edit it to tweak later and restart NFS pods. We don't have to re-spin NFS images for changes then. thoughts?

vendor/github.com/hpe-storage/common-host-libs/storageprovider/csp/container_storage_provider.go
shivamerla: This was fixed in latest common-host-libs. will update the package.

pkg/flavor/kubernetes/nfs.go
rgcostea: I guess I missed the check for one already existing.  Is that in the PR?

shivamerla: yes..we attempt create everytime, and if error is type IsAlreadyExists we return success.

rgcostea: Super! I missed that.  Thanks.

pkg/flavor/kubernetes/nfs.go
shivamerla: changed this to fetch from api-server pod.

```
time="2020-05-04T16:34:47Z" level=trace msg=">>>>> GetOrchestratorVersion" file="nfs.go:610"
time="2020-05-04T16:34:47Z" level=trace msg="obtained k8s version as v1.17.0" file="nfs.go:628"
time="2020-05-04T16:34:47Z" level=trace msg="<<<<< GetOrchestratorVersion" file="nfs.go:629"
```

PR #158, opened 2020-05-05
Require iscsi or multipath but do not fail in absence of one

jlarriba:
There can be situations where iscsi or multipath are not available. It is better not to fail on that scenarios thant requiring to have both in the OS.

### DIFF
--- a/cmd/csi-driver/csi-driver.go
+++ b/cmd/csi-driver/csi-driver.go
@@ -111,17 +111,28 @@ func csiCliHandler(cmd *cobra.Command) error {
 
 	if nodeService {
 		// perform conformance checks and service management
+		errors := 0
+
 		// configure iscsi
 		err = tunelinux.ConfigureIscsi()
 		if err != nil {
-			return fmt.Errorf("Unable to configure iscsid service, err %v", err.Error())
+			errors++
+			fmt.Printf("Unable to configure iscsid service, err %v", err.Error())
 		}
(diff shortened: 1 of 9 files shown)


cmd/csi-driver/csi-driver.go
```diff
@@ -111,17 +111,28 @@ func csiCliHandler(cmd *cobra.Command) error {
...
-			return fmt.Errorf("Unable to configure multipathd service, err %v", err.Error())
+			errors++
+			fmt.Printf("Unable to configure multipathd service, err %v", err.Error())
+		}
+		
+		// We want to be resilient in case the OS is not using either multipathd or iscsid
+		// but we need to use at least one
+		if errors > 1 {
+			return fmt.Errorf("Either iscsid or multipathd must be available")
 		}
```
rgcostea, line 134: Hi @jlarriba .  Thanks for the contribution.  But we actually need both iscsid and multipathd to be installed.

shivamerla: @jlarriba Can you close this PR based on @rgcostea comments above. We need both to be present on host for proper functioning of our CSI driver to attach block devices to worker nodes.

jlarriba: Yes, I forgot. Speaking with HPE, changes were made to the linux shared library that fixed the problem created by this issue.

PR #162, opened 2020-05-08
nfs: add tests

shivamerla:
* Problem:
  * unit tests for nfs support
* Implementation:
  * added few tests using fake k8s client
* Testing: using make test
* Review: gcostea, rkumar

### DIFF
--- a/pkg/flavor/kubernetes/nfs_test.go
+++ b/pkg/flavor/kubernetes/nfs_test.go
@@ -1,0 +1,133 @@
+package kubernetes
+
+import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


pkg/flavor/kubernetes/nfs_test.go
```diff
@@ -0,0 +1,133 @@
+package kubernetes
```
rgcostea, line 1: copyright missing

shivamerla: thanks..added.

PR #164, opened 2020-05-08
csi: update logic to fetch k8s server version

shivamerla:
* Problem:
  * use version package to obtain server version
* Implementation:
  * updated logic to use version package
* Testing: deployed and verified k8s version and parsing.
* Review: gcostea, rkumar

### DIFF
--- a/pkg/flavor/kubernetes/flavor.go
+++ b/pkg/flavor/kubernetes/flavor.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/version"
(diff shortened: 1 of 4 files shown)


pkg/flavor/kubernetes/flavor.go
```diff
@@ -574,3 +574,15 @@ func (flavor *Flavor) GetVolumePropertyOfPV(propertyName string, pvName string)
 	}
 	return "", nil
 }
+
+func (flavor *Flavor) GetOrchestratorVersion() (*version.Info, error) {
+	log.Tracef(">>>>> GetOrchestratorVersion")
+	defer log.Tracef("<<<<< GetOrchestratorVersion")
+
+	version, err := flavor.kubeClient.Discovery().ServerVersion()
```
raunakkumar, line 582: nit: maybe name this variable versionInfo. It looks like we have a version object here

shivamerla: sure..

PR #166, opened 2020-05-11
nfs: add service account while deploying nfs pod

shivamerla:
* Problem:
  * On openshift need to add securitycontextconstraint with serviceaccount
  * to allow creation of NFS pods just like CSI pods
* Implementation:
  * Add a method to create service account in namespace where we deploy NFS PVC/Pod.
  * cleanup method names as we don't need to export k8s specific methods outside of flavor package.
* Testing: added unit tests, tested on both OCP and K8s to verify service account, NFS pod deployment.
* Review: gcostea, rkumar


shivamerla:
> lgtm. you would be adding ClusterRole rules to perform CRUD on serviceAccounts in yaml or would that be handled only via operator?
yes, need to update yaml/templates to get/create service account as well.

PR #167, opened 2020-05-12
Updated ephemeral volume data and Added node scrubber for ephemeral pods

suneeth51:
- Update ephemeral nodPublish() to create separate data file 'ephemeral_data.json' under staging/targetpath.
- The data file 'ephemeral_data.json' stores the secret reference, pod info and volume details
- Added node scrubber support: (periodic task)
      -- Added CLI parameter 'scrubber-interval' which is set to 1800 seconds (default value).
      -- This scrubber task traverses all the ephemeral pods under '/var/lib/kubelet/pods' path on the node looking for 'ephemeral_data.json' and also 'deviceInfo.json'.
       -- From the above data, if the corresponding pod does not exist on the cluster, perform nodeUnpublishEphemeralVolume() to cleanup devices followed destroy volume from backend.

/* SAMPLE EPHEMERAL DATA FILE CONTENTS: 'ephemeral_data.json'
{
	"secret_ref" : {
		"namespace" : "kube-system",
		"name" : "nimble-secret"
	},
	"pod_data" : {
		"namespace" : "kube-system",
		"uid" : "daa1c178-9666-4afd-85fe-7906e1384c05",
		"name" : "my-pod-inline-mount-test"
	},
	"volume_handle" : "csi-f4f081da0a1ec68b65231cf29fe0d332330eaa136f0284d72c447f49ce665d87",
	"volume_id" : "0601a32bf8f45646a50000000000000000000005fd"
}
*/


pkg/driver/driver.go
```diff
@@ -157,6 +170,36 @@ func (driver *Driver) Stop(nodeService bool) error {
 	return nil
 }
 
+// StartScrubber starts the scrubber period task
+func (driver *Driver) StartScrubber(nodeService bool, scrubberInterval int64) error {
+	if nodeService {
+		log.Tracef("Scheduling scrubber task to run every %v seconds", scrubberInterval)
```
raunakkumar, line 176: nit: %s

pkg/driver/driver.go
```diff
@@ -645,4 +688,96 @@ func (driver *Driver) IsNFSResourceRequest(parameters map[string]string) bool {
...
+	exists, isDir, err := util.FileExists(podsDirPath)
+	if err != nil {
+		return err
+	}
+	if !exists || !isDir {
+		return fmt.Errorf("Path [%s] does not exist or not a directory", podsDirPath)
+	}
+
+	// Fetch all ephemeral pods and populate it
+	//ephemeralPods := map[string][]*StagingDeviceEphemeralData{}
```
rgcostea, line 717: Can this line be removed?

pkg/driver/driver.go
```diff
@@ -645,4 +688,96 @@ func (driver *Driver) IsNFSResourceRequest(parameters map[string]string) bool {
...
+				}
+
+				// Add volumePath to the pod list
+				volHandleTargetPath := &VolumeHandleTargetPath{
+					VolumeHandle: ephemeralData.VolumeHandle,
+					TargetPath:   targetPath,
+				}
+				log.Tracef("Adding ephemeral pod/volume scrubbing, [POD: %s], [VolumeHandle: %s], [VolumeID: %s], [TargetPath: %s]",
+					ephemeralData.PodData.UID, ephemeralData.VolumeHandle, ephemeralData.VolumeID, targetPath)
+				ephemeralPods[ephemeralData.PodData.UID] = append(ephemeralPods[ephemeralData.PodData.UID], volHandleTargetPath)
```
raunakkumar, line 744: check if the key already exists?

cmd/csi-driver/csi-driver.go
```diff
@@ -26,15 +26,17 @@ const (
 	csiControllerLogFile = "/var/log/hpe-csi-controller.log"
 	csiNodeLogFile       = "/var/log/hpe-csi-node.log"
 	csiEndpoint          = "unix:///var/lib/kubelet/csi.hpe.com/csi.sock"
+	csiScrubberInterval  = 1800 // Default value is 1800 seconds or 30 minutes
```
rgcostea, line 29: Should we make it clear that this is an inline volume scrubber?  Just calling it csiScrubberInterval can lead to confusion... what exactly are we scrubbing?

cmd/csi-driver/csi-driver.go
```diff
@@ -71,6 +73,7 @@ func init() {
 	RootCmd.PersistentFlags().StringVarP(&dbServer, "dbserver", "s", "", "Database server for the CSI driver")
 	RootCmd.PersistentFlags().StringVarP(&dbPort, "dbport", "p", etcd.DefaultPort, "Database server port for the CSI driver")
 	RootCmd.PersistentFlags().BoolP("node-service", "", false, "CSI node-plugin")
+	RootCmd.PersistentFlags().Int64Var(&scrubberInterval, "scrubber-interval", csiScrubberInterval, "Scrubber interval in seconds for the CSI driver")
```
rgcostea, line 76: Same comment here: Maybe calling it inlineVolumeScrubberInterval would make this more obvious.

cmd/csi-driver/csi-driver.go
suneeth51: I thought of keeping it generic so that any future scrubbing tasks can be piggy backed. However, it makes sense to keep to specific. Here we are scrubbing ephemeral pods

rgcostea: I'm curious though.  Why do we feel that we need to expose this to the user via our flag variables?  Is this something we expect them to change?

pkg/driver/driver.go
```diff
@@ -157,6 +170,36 @@ func (driver *Driver) Stop(nodeService bool) error {
...
+		tick := time.NewTicker(time.Duration(scrubberInterval) * time.Second)
+		done := make(chan bool)
+		go func() {
+			// Cleanup ephemeral orphan volumes
+			driver.ScrubberTask(time.Now())
+			for {
+				select {
+				case t := <-tick.C:
+					driver.ScrubberTask(t)
+				case <-done:
```
shivamerla, line 186: Where is this channel closed? on stop you are missing it?

pkg/driver/driver.go
raunakkumar: do we expect more than 1 VolumeHandleTargetPath for a single podID? I had thought we have 1-1 mapping between pod and volume. In case of failover too a new pod is created for which we create a new volume.

suneeth51: If single pod has two volumes, then we will have two target paths, one for each volume.
ex:
/var/lib/kubelet/pods/b627d92c-022b-461f-84a3-2163b033bd6f/volumes/kubernetes.io~csi/my-volume-vols-1/mount
/var/lib/kubelet/pods/b627d92c-022b-461f-84a3-2163b033bd6f/volumes/kubernetes.io~csi/my-volume-vols-2/mount

cmd/csi-driver/csi-driver.go
suneeth51: The idea is to keep it configurable so that we can update the deployment yamls to modify it accordingly to control of frequency of scrubbing. Scrubber does traverses the pods directory path recursively `/var/lib/kubelet/pods/`.

rgcostea: I'm not questioning why we allow it to be configurable.  I'm just wondering why we think it needs to be user facing.  We can treat it like we do the volume limit and make it an environment variable.

suneeth51: Yes, I think that would be more ideal. Let me fix it that way instead?

pkg/driver/driver.go
suneeth51: Yes, it got missed. I will fix this.

PR #168, opened 2020-05-13
nfs-monitor: added nfs pod monitor to handle node partition cases

shivamerla:
* Problem:
  * when a node running NFS pod is partitioned, k8s doesn't move them to
  * different node and manual intervention is required.
* Implementation:
  * added nfs pod monitor to periodically check nfs pods running on lost nodes
  * and once podEvictionTimeout is expired, will force delete the volume attachment
  * and pod to let k8s schedule them on good nodes without getting multi-attach errors.
* Testing: tested with multiple scenarios of node partitioning and verified pod is started
  * without manual intervention on different node. pending integration/unit tests for VA deletions.


pkg/flavor/kubernetes/monitor.go
```diff
@@ -0,0 +1,222 @@
...
+const (
+	defaultIntervalSec   = 30
+	minimumIntervalSec   = 15
+	nodeUnreachableTaint = "node.kubernetes.io/unreachable"
+	nodeLostCondition    = "NodeLost"
+)
+
+// Monitor Pods running on un-reachable nodes
+type Monitor struct {
+	IntervalSec int64
```
raunakkumar, line 27: nit: IntervalSecond

cmd/csi-driver/csi-driver.go
```diff
@@ -30,11 +31,13 @@ const (
...
-	dbServer   string
-	dbPort     string
-	flavorName string
+	name               string
+	endpoint           string
+	dbServer           string
+	dbPort             string
+	flavorName         string
+	nfsMonitor         bool
+	nfsMonitorInterval string
```
rgcostea, line 40: This all looks great.  But I have to ask, why limit this to monitoring only for nfs pods?  The monitor function below can allow any label to be specified.  Why not pull that out into a separate library that can be extended for other purposes?

shivamerla: Yes, i was little cautious to make it generic as clean I/O shutdown is still an issue on failed node and thought coverage with NFS should provide good data. I can make these generic and get labels through ENV.

pkg/driver/driver.go
```diff
@@ -41,6 +41,7 @@ type Driver struct {
 	storageProviders map[string]storageprovider.StorageProvider
 	flavor           flavor.Flavor
 	grpc             NonBlockingGRPCServer
+	nfsPodMonitor    *kubernetes.Monitor
```
rgcostea, line 44: This is now starting to bleed k8s constructs into our driver.  The driver should remain agnostic.  We should continue to put this type of stuff into our flavor.

pkg/flavor/kubernetes/flavor.go
```diff
@@ -49,10 +50,11 @@ type Flavor struct {
 	claimStopChan chan struct{}
 
 	eventRecorder record.EventRecorder
+	nfsMonitor    *Monitor
 }
 
 // NewKubernetesFlavor creates a new k8s flavored CSI driver
-func NewKubernetesFlavor(nodeService bool) (*Flavor, error) {
+func NewKubernetesFlavor(nodeService bool, nfsMonitor bool, nfsMonitorInterval int64) (*Flavor, error) {
```
rgcostea, line 57: It might be cleaner to just do this in a separate function:

func ConfigureMonitor(type nfs|minio, monitorInterval)

Thus allowing the client to determine which type and interval to apply.

pkg/driver/driver.go
```diff
@@ -136,24 +137,30 @@ func getDBClient(server string, port string) (*etcd.DBClient, error) {
...
+func (driver *Driver) Start(nodeService bool, nfsMonitor bool) error {
 	go func() {
 		driver.grpc = NewNonBlockingGRPCServer()
 		if nodeService {
 			driver.grpc.Start(driver.endpoint, driver, nil, driver)
 		} else {
 			driver.grpc.Start(driver.endpoint, driver, driver, nil)
+			if nfsMonitor {
+				driver.flavor.StartNFSMonitor()
+			}
```
rgcostea, line 149: Once you change the interface to configureMonitor, this can change to StartMonitors() which will start all registered monitors.

shivamerla: @rgcostea do you think its better to pull this into a separate sidecar instead?

rgcostea: I'm not sure that is necessarily better as that includes the overhead of building and managing another container.

shivamerla: ok.let me pull this into separate package then with above changes.

pkg/driver/driver.go
shivamerla: ah..looks like i forgot to clean this up. This is not used and moved to Flavor instead.
```// Flavor of the CSI driver
type Flavor struct {
	crdClient  *crd_client.Clientset
	nodeName   string
	kubeClient kubernetes.Interface

	claimInformer cache.SharedIndexInformer
	claimIndexer  cache.Indexer
	claimStopChan chan struct{}

	eventRecorder record.EventRecorder
	nfsMonitor    *Monitor
}```

File: pkg/monitor/monitor_test.go [hunk 6]
```diff
@@ -0,0 +1,38 @@
+package monitor
```
rgcostea, line 1: missing copyright

shivamerla: fixed...thanks.

PR #171, opened 2020-05-17
Updated inline ephemeral nodePublish workflow

suneeth51:
Signed-off-by: Suneeth Kumar Byadara Halli <suneethkumar.byadarahalli@hpe.com>

- Refactored the ephemeral volume creation/publish workflow

### DIFF
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -60,8 +60,11 @@ const (
 	// driverModeKey stored in the vol_data.json file
 	driverModeKey = "driverMode"
 
-	// ephemeralDriverMode for ephmeral inline volume
-	ephemeralDriverMode = "ephemeral"
+	// volumeLifecycleMode stored in the vol_data.json file
+	volumeLifecycleModeKey = "volumeLifecycleMode"
+
+	// ephemeralVolumeMode for ephmeral inline volume
+	ephemeralVolumeMode = "ephemeral"
 
 	// Default scrubber interval for Ephemeral inline volumes on the node.
 	defaultInlineVolumeScrubberInterval = 3600 // Default value is 3600 seconds or 1 Hour
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -63,7 +63,7 @@ const (
 	// volumeLifecycleMode stored in the vol_data.json file
 	volumeLifecycleModeKey = "volumeLifecycleMode"
 
-	// ephemeralVolumeMode for ephmeral inline volume
+	// ephemeralVolumeMode for ephemeral inline volume
 	ephemeralVolumeMode = "ephemeral"
 
 	// Default scrubber interval for Ephemeral inline volumes on the node.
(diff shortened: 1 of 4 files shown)


pkg/driver/constants.go
```diff
@@ -60,8 +60,11 @@ const (
 	// driverModeKey stored in the vol_data.json file
 	driverModeKey = "driverMode"
 
-	// ephemeralDriverMode for ephmeral inline volume
-	ephemeralDriverMode = "ephemeral"
+	// volumeLifecycleMode stored in the vol_data.json file
+	volumeLifecycleModeKey = "volumeLifecycleMode"
+
+	// ephemeralVolumeMode for ephmeral inline volume
```
rgcostea, line 66: typo: ephmeral

pkg/driver/node_server.go
```diff
@@ -708,89 +708,36 @@ func (driver *Driver) NodePublishVolume(ctx context.Context, request *csi.NodePu
...
-						fmt.Sprintf("%s, %s", err.Error(), rbErr.Error()))
-				}
-				log.Infof("Cleanup/Rollback of ephemeral volume %s successful", request.VolumeId)
-				return nil, err
-			}
+		err := driver.nodePublishEphemeralVolume(
+			request.VolumeId, // Ephemeral CSI Volume Handle
+			volumeName,
+			stagingTargetPath,
+			request.TargetPath,
```
shivamerla, line 729: may be better to just pass entire request object itself?

pkg/driver/node_server.go
```diff
@@ -1028,152 +1015,125 @@ func (driver *Driver) nodePublishEphemeralVolume(
...
+			return status.Error(codes.Internal,
+				fmt.Sprintf("%s, %s", err.Error(), rbErr.Error()))
 		}
+		log.Infof("Rollback successful for ephemeral volume %s", volumeName)
+		return err
 	}
+	return nil
+}
+
+func (driver *Driver) createPublishEphemeralVolume(
```
shivamerla, line 1084: is there a reason to tie these two together? can create and publish volume be still kept separate?

pkg/driver/constants.go
suneeth51: Agreed, I will fix

pkg/driver/node_server.go
suneeth51: `createPublishEphemeralVolume` is just another separate wrapper function which internally calls create(), controlPublish(), nodeStage() and nodePublish() separately. There are grouped together so that we can rollback if any failure occurs within those intermediate steps.

shivamerla: got it. thanks.

pkg/driver/node_server.go
suneeth51: Some attributes like 'volumeName' from 'volumeID' and 'stagingTargetPath' from 'targetPath' are derived/constructed from the request object before passing down to this function and other functions. so, all these fields intact and common they are passed separately to avoid another interpretation with that called function.

shivamerla: ok..thanks.

PR #172, opened 2020-05-18
nfs: add conformance checks for nfs  packages

shivamerla:
* Problem:
  * nfs-utils/nfs-common packages are not always present on node
* Implementation:
  * On redhat/centos versions check for nfs-utils package
  * On Ubuntu check for nfs-common package
  * enable NFS client services on node(nfs-utils.service on both redhat/ubuntu)
  * For CoreOS skip all installation as they are immutable.
* Testing: verified running the script on both redhat and ubuntu

### DIFF
--- a/cmd/csi-driver/conform/hpe-storage-node.sh
+++ b/cmd/csi-driver/conform/hpe-storage-node.sh
@@ -3,7 +3,7 @@
 exit_on_error() {
     exit_code=$1
     if [ $exit_code -ne 0 ]; then
-        >&2 echo "command failed with exit code ${exit_code}."
+        echo "command failed with exit code ${exit_code}."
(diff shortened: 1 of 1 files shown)


cmd/csi-driver/conform/hpe-storage-node.sh
```diff
@@ -68,9 +68,9 @@ elif [ "$CONFORM_TO" = "redhat" ]; then
     # Install nfs client packages
     if [ ! -f /sbin/mount.nfs4 ]; then
         yum -y install nfs-utils
-        exit_on_error $?
```
raunakkumar: should we rather check with rpm -qa if nfs-utils is already there?

PR #174, opened 2020-05-20
nfs: misc nfs fixes

shivamerla:
* Problem:
  * 1. Domain name used by nfs-server by default(localdomain) causes conflict
  * when NFS clients require host domain name for mounts. This causes nobody:nobody files.
  * 2. Default NFS version is 3 in some Ubuntu distros, causing RPC procedure
  * not available errors
  * 3. Foreground deletion is used to cleanup resources, which causes timeout
  * of RWX PV deletion, but when retry is attempted, since underlying RWO PV
  * is deleted by then, we can't figureout RWX PV to be of NFS type, causing stale PV's.
  * 4. Default GracePeriod for NFS server to be idle after startup is 90s. This causes
  * initial I/O to be hung soon after app pods are started.
  * 5. ConfigMap is not displayed in pretty format during edit
* Implementation:
  * 1. Domain name is added to ganesha.conf using dnsdomainname, allowing NFS clients
  * to mount correctly where idmap is disabled.
  * 2. Default vers=4 is set during mount as NFS server is always started with 4.0 version as default.
  * It can be overridden by user using configmap and SC param nfsMountOptions
  * 3. Changed cleanup logic to background, so timeouts wont happen with RWX PV.
  * In the background, these deletions can take time and k8s takes care of that.
  * 4. Start NFS server with GraceLess option. Seems like general practice in container mode.
  * 5. Format configmap with individual strings rather than using back ticks. This allows user to easily edit it later.
* Testing: tested with deployments on Ubuntu/RHEL/CoreOS.
* Review: gcostea, rkumar

### DIFF
--- a/pkg/flavor/kubernetes/nfs.go
+++ b/pkg/flavor/kubernetes/nfs.go
@@ -58,7 +58,7 @@ type NFSSpec struct {
 }
 
 // CreateNFSVolume creates nfs volume abstracting underlying nfs pvc, deployment and service
-func (flavor *Flavor) CreateNFSVolume(chapiDriver chapi.Driver, pvName string, reqVolSize int64, parameters map[string]string, volumeContentSource *csi.VolumeContentSource) (nfsVolume *csi.Volume, rollback bool, err error) {
+func (flavor *Flavor) CreateNFSVolume(pvName string, reqVolSize int64, parameters map[string]string, volumeContentSource *csi.VolumeContentSource) (nfsVolume *csi.Volume, rollback bool, err error) {
 	log.Tracef(">>>>> CreateNFSVolume with %s", pvName)
 	defer log.Tracef("<<<<< CreateNFSVolume")
 
@@ -111,17 +111,12 @@ func (flavor *Flavor) CreateNFSVolume(chapiDriver chapi.Driver, pvName string, r
 		return nil, true, err
 	}
 
-	// obtain an array of  {hostname, domainname}
-	hostNameAndDomain, err := chapiDriver.GetHostNameAndDomain()
+	nfsHostDomain, err := flavor.getNFSHostDomain()
 	if err != nil {
-		return nil, true, fmt.Errorf("Failed to obtain host name and domain to provision NFS volume, %s", err.Error())
+		return nil, true, err
 	}
-	if len(hostNameAndDomain) != 2 || hostNameAndDomain[1] == "unknown" {
-		return nil, true, fmt.Errorf("Unable to obtain host domain name to provision NFS volume %s", pvName)
-	}
-	log.Tracef("Host domain name obtained as %s", hostNameAndDomain[1])
 	// create nfs configmap
-	err = flavor.createNFSConfigMap(nfsResourceNamespace, hostNameAndDomain[1])
+	err = flavor.createNFSConfigMap(nfsResourceNamespace, nfsHostDomain)
 	if err != nil {
 		flavor.eventRecorder.Event(claim, core_v1.EventTypeWarning, "ProvisionStorage", err.Error())
 		return nil, true, err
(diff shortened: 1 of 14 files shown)


pkg/flavor/kubernetes/nfs.go
```diff
@@ -60,7 +58,7 @@ type NFSSpec struct {
 }
 
 // CreateNFSVolume creates nfs volume abstracting underlying nfs pvc, deployment and service
-func (flavor *Flavor) CreateNFSVolume(pvName string, reqVolSize int64, parameters map[string]string, volumeContentSource *csi.VolumeContentSource) (nfsVolume *csi.Volume, rollback bool, err error) {
+func (flavor *Flavor) CreateNFSVolume(chapiDriver chapi.Driver, pvName string, reqVolSize int64, parameters map[string]string, volumeContentSource *csi.VolumeContentSource) (nfsVolume *csi.Volume, rollback bool, err error) {
```
rgcostea, line 61: Rather than passing the chapi Driver around in different functions, can we instantiate an instance of it on the flavor when it get initialized?  Or give it the reference from the driver?

pkg/flavor/kubernetes/nfs.go
```diff
@@ -179,32 +186,15 @@ func (flavor *Flavor) createServiceAccount(nfsNamespace string) error {
...
-  SecType = "sys";
-  FSAL {
-	  Name = VFS;
-  }
-}`
+	// Format for pretty display with configmap
+	nfsCoreParamBlock := "NFS_Core_Param" + "\n" + "{" + "\n" + "NFS_Protocols= 4;" + "\n" + "NFS_Port = 2049;" + "\n" + "fsid_device = false;" + "\n" + "}" + "\n"
+	nfs4Block := "NFSv4" + "\n" + "{" + "\n" + "Graceless = true;" + "\n" + "UseGetpwnam = true;" + "\n" + "DomainName = " + hostDomain + ";" + "\n" + "}" + "\n"
+	exportBlock := "EXPORT" + "\n" + "{" + "\n" + "Export_Id = 716;" + "\n" + "Path = /export;" + "\n" + "Pseudo = /export;" + "\n" + "Access_Type = RW;" + "\n" + "Squash = No_Root_Squash;" + "\n" + "Transports = TCP;" + "\n" + "Protocols = 4;" + "\n" + "SecType = \"sys\";" + "\n" + "FSAL" + "\n" + "{" + "\n" + "Name = VFS;" + "\n" + "}" + "\n" + "}" + "\n"
+	nfsGaneshaConfig := nfsCoreParamBlock + nfs4Block + exportBlock
```
rgcostea, line 197: It looks prettier the way it was... at least in code :)

pkg/flavor/kubernetes/nfs.go
shivamerla: done..thanks.

pkg/flavor/kubernetes/nfs.go
shivamerla: i had an extra tab in the original code looks like, so during edit of configmap using kubectl, it was escaping all new lines. Now its fixed, so reverted back.

pkg/flavor/kubernetes/nfs.go
```diff
@@ -359,6 +368,7 @@ func (flavor *Flavor) HandleNFSNodePublish(chapiDriver chapi.Driver, req *csi.No
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	chapiDriver := &chapi.LinuxDriver{}
```
rgcostea, line 371: What if this is running on Windows?  Should we check the OS type?

Not that we support that now.  I'm just asking.

pkg/flavor/kubernetes/nfs.go
```diff
@@ -396,6 +406,23 @@ func (flavor *Flavor) GetNFSVolumeID(volumeID string) (string, error) {
 	return pv.Spec.PersistentVolumeSource.CSI.VolumeHandle, nil
 }
 
+func (flavor *Flavor) getNFSHostDomain() (string, error) {
+	log.Tracef(">>>>> getNFSHostDomain")
+	defer log.Tracef("<<<<< getNFSHostDomain")
+
+	// obtain an array of  {hostname, domainname}
+	chapiDriver := &chapi.LinuxDriver{}
```
rgcostea, line 414: I meant passing in the reference created in the driver?  Why do we want to construct this in each function?

shivamerla: ```		driver.flavor.chapiDriver = driver.chapiDriver``` this way so that we can reference using flavor?

pkg/flavor/kubernetes/nfs.go
shivamerla: sure, if we use the same reference created in driver, that should be handled using getDriver() call.

rgcostea: Exactly.

pkg/flavor/kubernetes/nfs.go
rgcostea: Ya.  Then you just have a chapiDriver reference in flavor that was initialized in driver_<ostype>.go.

shivamerla: understood..much cleaner..fixed it.

PR #176, opened 2020-05-28
nfs: increase wait time for deployment to 5min

shivamerla:
* Problem:
  * If app pods and NFS pods are started simultaneously, then
  * NFS mounts are slow initially and can take more than 2 mins to start.
  * Current wait time of 2min, will cause un-necessary rollbacks and retries.
* Implementation:
  * Change minimum wait interval to 5min to avoid un-necessary rollback and retries
  * waiting for NFS deployment to be ready.
* Testing: tested with 20 NFS pods and 80 application pods.
* Review: rkumar

### DIFF
--- a/pkg/flavor/kubernetes/nfs.go
+++ b/pkg/flavor/kubernetes/nfs.go
@@ -26,8 +26,8 @@ const (
 	defaultNFSNamespace = "hpe-nfs"
 	defaultNFSImage     = "hpestorage/nfs-provisioner:2.8.3-4"
 
-	creationInterval           = 150 // 300s with sleep interval of 2s
-	creationDelay              = 2 * time.Second
+	creationInterval           = 60 // 300s with sleep interval of 5s
+	creationDelay              = 5 * time.Second
 	defaultExportPath          = "/export"
 	nfsResourceLimitsCPUKey    = "nfsResourceLimitsCpuM"
 	nfsResourceLimitsMemoryKey = "nfsResourceLimitsMemoryMi"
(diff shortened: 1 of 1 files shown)


pkg/flavor/kubernetes/nfs.go
```diff
@@ -965,6 +964,9 @@ func (flavor *Flavor) waitForPVCCreation(claimName, nfsNamespace string) error {
 			log.Infof("pvc %s still not bound to pv, current state %s. waiting(try %d)...", claimName, claim.Status.Phase, i+1)
 			time.Sleep(sleepTime)
 			continue
+		} else if !claim.ObjectMeta.DeletionTimestamp.IsZero() {
+			log.Errorf("existing pvc %s is under deletion, return error to retry later...", claimName)
```
raunakkumar, line 968: log this as debug/trace as it's unrelated to the error below.

pkg/flavor/kubernetes/nfs.go
```diff
@@ -26,7 +26,7 @@ const (
 	defaultNFSNamespace = "hpe-nfs"
 	defaultNFSImage     = "hpestorage/nfs-provisioner:2.8.3-4"
 
-	creationInterval           = 60 // 120s with sleep interval of 2s
+	creationInterval           = 150 // 300s with sleep interval of 2s
```
raunakkumar, line 29: instead of 150 tries, would it be better to sleep for 5 seconds instead of 2 ?

pkg/flavor/kubernetes/nfs.go
shivamerla: sure, changed it to log.warnf() as its very rare and useful for debugging without trace levels.

pkg/flavor/kubernetes/nfs.go
shivamerla: sounds good as deployment usually takes more than 5s initially to start. changed it to 5s interval.

PR #178, opened 2020-05-29
nfs: Assume NFS resources for RWX/ROX volumes

shivamerla:
* Problem:
  * Currently when RWX/ROX volumes are created without nfsResources flag
  * we allow block volume creation, without any NFS resources.
  * As this volume can be attached to multiple nodes/apps, without distributed
  * filesystem, this will cause data inconsistencies.
* Implementation:
  * Assume nfsResources are requested with RWX/ROX access types. When backend
  * itself supports NFS, then we need to get serverIP etc from sc params and
  * when that support is added, we can change this assumption along with other changes
  * to bypass node stage etc.
* Testing: creation of RWX/ROX volumes without any additional NFS flags.
* Review: gcostea, rkumar

### DIFF
--- a/pkg/flavor/kubernetes/nfs.go
+++ b/pkg/flavor/kubernetes/nfs.go
@@ -904,7 +904,8 @@ func (flavor *Flavor) deleteNFSDeployment(name string, nfsNamespace string) erro
 	defer log.Tracef("<<<<< deleteNFSDeployment")
 
 	propagation := meta_v1.DeletePropagationBackground
-	options := &meta_v1.DeleteOptions{PropagationPolicy: &propagation}
+	gracePeriod := int64(5)
+	options := &meta_v1.DeleteOptions{PropagationPolicy: &propagation, GracePeriodSeconds: &gracePeriod}
 
 	err := flavor.kubeClient.AppsV1().Deployments(nfsNamespace).Delete(name, options)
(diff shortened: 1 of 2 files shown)


pkg/flavor/kubernetes/nfs.go
```diff
@@ -904,7 +904,8 @@ func (flavor *Flavor) deleteNFSDeployment(name string, nfsNamespace string) erro
 	defer log.Tracef("<<<<< deleteNFSDeployment")
 
 	propagation := meta_v1.DeletePropagationBackground
-	options := &meta_v1.DeleteOptions{PropagationPolicy: &propagation}
+	gracePeriod := int64(5)
```
raunakkumar, line 907: nit: you can directly pass int64(5) in the DeleteOptions

shivamerla: @raunakkumar we need to pass this as pointer hence separate variable.

shivamerla:
Error:
```
  Warning  ProvisioningFailed    4s (x4 over 11s)  csi.hpe.com_skrishna-u16k8s-3_ca0010ed-51e9-4f9d-be08-4122f0dd0d40  failed to provision volume with StorageClass "hpe-standard": rpc error: code = InvalidArgument desc = StorageClass parameter nfsResources=true is missing for creation of volumes with multi-node access```

PR #188, opened 2020-06-29
Accomodate scanning for multiple lun id

wdurairaj:
This PR has a dependency on common-host-libs PR -- https://github.com/hpe-storage/common-host-libs/pull/126 
Please merge the common-host-libs PR first, and then this.

- Requirement

    - Current in 3PAR/Primera CSP we are adding requirement for Peer persistence based Replication, where we need to create host and vlun's for both the Primary (source) array and the Target (secondary) array, so that the multipath device mapping will have paths from active/passive arrays. But, the publish volume response from CSP to CSI has currently support for only returning a single lun id. The change here is to return a object called SecondaryArrayDetails which will have list of object PeerDetails which has lun id, discovery ip, target names etc. One PeerDetail will be created per secondary backend.

- Approach

Changes are done to find, if the multiple luns are passed based on the number of PeerDetails in SecondaryArrayDetails struct. iscsi login , scanning all takes care of reading this object, and invokes appropriate actions.

Signed-off-by: William Durairaj <w_durairaj@yahoo.com>

### DIFF
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -41,7 +41,7 @@ const (
 	targetScopeKey    = "targetScope"
 	lunIDKey          = "lunId"
 
-	secondaryArrayDetailsKey = "secondArrayDetailsKey"
+	secondaryArrayDetailsKey = "secondaryArrayDetailsKey"
 
 	discoveryIPsKey       = "discoveryIps"
 	chapUsernameKey       = "chapUsername"
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -43,10 +43,6 @@ const (
 
 	secondaryArrayDetailsKey = "secondArrayDetailsKey"
 
-	secondaryLunIDKey       = "secondaryLunIdKey"
-	secondaryTargetNamesKey = "secondaryTargetNamesKey"
-	secondaryDiscoveryIpKey = "secondaryDiscoveryIpKey"
-
 	discoveryIPsKey       = "discoveryIps"
 	chapUsernameKey       = "chapUsername"
 	chapPasswordKey       = "chapPassword"
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -41,6 +41,8 @@ const (
 	targetScopeKey    = "targetScope"
 	lunIDKey          = "lunId"
 
+	secondaryArrayDetailsKey = "secondArrayDetailsKey"
+
 	secondaryLunIDKey       = "secondaryLunIdKey"
 	secondaryTargetNamesKey = "secondaryTargetNamesKey"
 	secondaryDiscoveryIpKey = "secondaryDiscoveryIpKey"
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -40,6 +40,7 @@ const (
 	targetNamesKey        = "targetNames"
 	targetScopeKey        = "targetScope"
 	lunIDKey              = "lunId"
+	peerLunIDs            = "peer_lun_ids"
 	discoveryIPsKey       = "discoveryIps"
 	chapUsernameKey       = "chapUsername"
 	chapPasswordKey       = "chapPassword"
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -35,12 +35,16 @@ const (
 	targetScopeGroup = "group"
 
 	// Volume Publish params
-	serialNumberKey       = "serialNumber"
-	accessProtocolKey     = "accessProtocol"
-	targetNamesKey        = "targetNames"
-	targetScopeKey        = "targetScope"
-	lunIDKey              = "lunId"
-	peerLunIDs            = "peer_lun_ids"
+	serialNumberKey   = "serialNumber"
+	accessProtocolKey = "accessProtocol"
+	targetNamesKey    = "targetNames"
+	targetScopeKey    = "targetScope"
+	lunIDKey          = "lunId"
+
+	secondaryLunIDKey       = "secondaryLunIdKey"
+	secondaryTargetNamesKey = "secondaryTargetNamesKey"
+	secondaryDiscoveryIpKey = "secondaryDiscoveryIpKey"
+
 	discoveryIPsKey       = "discoveryIps"
 	chapUsernameKey       = "chapUsername"
 	chapPasswordKey       = "chapPassword"
--- a/pkg/driver/controller_server.go
+++ b/pkg/driver/controller_server.go
@@ -867,12 +867,12 @@ func (driver *Driver) controllerPublishVolume(
 	publishContext[lunIDKey] = strconv.Itoa(int(publishInfo.AccessInfo.BlockDeviceAccessInfo.LunID))
 
 	// Start of population of target array details
-	log.Tracef("\n PUBLISH INFO ::::: %v", publishInfo)
-
 	if publishInfo.AccessInfo.BlockDeviceAccessInfo.SecondaryBackendDetails.PeerArrayDetails != nil {
 		secondaryArrayMarshalledStr, err := json.Marshal(&publishInfo.AccessInfo.BlockDeviceAccessInfo.SecondaryBackendDetails)
 		if err != nil {
 			log.Errorf("Error in marshalling secondary details %s", err.Error())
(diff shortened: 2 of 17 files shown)


shivamerla:
@wdurairaj few more things to consider as per our discussion:

1. Create a new type in publishinfo with required info to stage volume with secondary array (list of secondary array publish info: {iqn's, lun-id, discovery-ips}, this will be populated only if replication is setup.
2. Rescan/Remap changes are to handle only when secondary array info is not nil.
3. Fix multipath.conf for 3Par/Primera to use "group_by_prio" and "1 alua" handler as we now have stand-by paths which should not be grouped with active ones.

wdurairaj:
> @wdurairaj few more things to consider as per our discussion:
> 
> 1. Create a new type in publishinfo with required info to stage volume with secondary array (list of secondary array publish info: {iqn's, lun-id, discovery-ips}, this will be populated only if replication is setup.
> 2. Rescan/Remap changes are to handle only when secondary array info is not nil.
> 3. Fix multipath.conf for 3Par/Primera to use "group_by_prio" and "1 alua" handler as we now have stand-by paths which should not be grouped with active ones.

For point 3. i will work further.. Please check if the approach for point 1,2 is fine.

shivamerla:
> > @wdurairaj few more things to consider as per our discussion:
> > 
> > 1. Create a new type in publishinfo with required info to stage volume with secondary array (list of secondary array publish info: {iqn's, lun-id, discovery-ips}, this will be populated only if replication is setup.
> > 2. Rescan/Remap changes are to handle only when secondary array info is not nil.
> > 3. Fix multipath.conf for 3Par/Primera to use "group_by_prio" and "1 alua" handler as we now have stand-by paths which should not be grouped with active ones.
> 
> For point 3. i will work further.. Please check if the approach for point 1,2 is fine.
@wdurairaj me and George have discussed on this. We feel its better to add a new struct like SecondaryLunInfo which will contain all information required to publish/stage device with paths to secondary array volume. So, instead of PeerLunID's can you add this and populate within CSP? So, all logic will handle cases only when secondary volume info is not empty.

pkg/driver/controller_server.go
```diff
@@ -864,6 +865,23 @@ func (driver *Driver) controllerPublishVolume(
 	publishContext[targetNamesKey] = strings.Join(publishInfo.AccessInfo.BlockDeviceAccessInfo.TargetNames, ",")
 	publishContext[targetScopeKey] = requestedTargetScope
 	publishContext[lunIDKey] = strconv.Itoa(int(publishInfo.AccessInfo.BlockDeviceAccessInfo.LunID))
+
+	// Start of population of target array details
+	log.Tracef("\n PUBLISH INFO ::::: %v", publishInfo)
+	numberOfSecondaryBackends := len(publishInfo.AccessInfo.BlockDeviceAccessInfo.SecondaryBackendDetails.PeerArrayDetails)
```
shivamerla, line 871: do we need the count, can we just check if SecondaryBackendDetails not being nil?

wdurairaj: Done

pkg/driver/controller_server.go
```diff
@@ -864,6 +865,20 @@ func (driver *Driver) controllerPublishVolume(
 	publishContext[targetNamesKey] = strings.Join(publishInfo.AccessInfo.BlockDeviceAccessInfo.TargetNames, ",")
 	publishContext[targetScopeKey] = requestedTargetScope
 	publishContext[lunIDKey] = strconv.Itoa(int(publishInfo.AccessInfo.BlockDeviceAccessInfo.LunID))
+
+	// Start of population of target array details
+	log.Tracef("\n PUBLISH INFO ::::: %v", publishInfo)
+
+	if publishInfo.AccessInfo.BlockDeviceAccessInfo.SecondaryBackendDetails.PeerArrayDetails != nil {
```
shivamerla, line 872: Can you change this to ```if publishInfo.AccessInfo.BlockDeviceAccessInfo.SecondaryBackendDetails != nil``` as it will cause panic otherwise.

pkg/driver/controller_server.go
```diff
@@ -864,6 +865,20 @@ func (driver *Driver) controllerPublishVolume(
 	publishContext[targetNamesKey] = strings.Join(publishInfo.AccessInfo.BlockDeviceAccessInfo.TargetNames, ",")
 	publishContext[targetScopeKey] = requestedTargetScope
 	publishContext[lunIDKey] = strconv.Itoa(int(publishInfo.AccessInfo.BlockDeviceAccessInfo.LunID))
+
+	// Start of population of target array details
+	log.Tracef("\n PUBLISH INFO ::::: %v", publishInfo)
```
shivamerla, line 870: We already log publishInfo at line 853. can you remove this.

pkg/driver/controller_server.go
```diff
@@ -864,6 +865,20 @@ func (driver *Driver) controllerPublishVolume(
...
+
+	if publishInfo.AccessInfo.BlockDeviceAccessInfo.SecondaryBackendDetails.PeerArrayDetails != nil {
+		secondaryArrayMarshalledStr, err := json.Marshal(&publishInfo.AccessInfo.BlockDeviceAccessInfo.SecondaryBackendDetails)
+		if err != nil {
+			log.Errorf("Error in marshalling secondary details %s", err.Error())
```
shivamerla, line 875: i think we should fail publish request, if there is an error here as we cannot connect to secondary array without these in this case.

pkg/driver/controller_server.go
wdurairaj: I could'nt add above IF condition with != nil , since the compiler complains like
```
pkg/driver/controller_server.go:870:60: cannot convert nil to type model.SecondaryBackendDetails
```
I have confirmed the code changes with or without `SecondaryBackendDetails` populated via the CSP

pkg/driver/controller_server.go
wdurairaj: I added code to return nil, error response.

pkg/driver/controller_server.go
wdurairaj: Removed the logger line

pkg/driver/controller_server.go
shivamerla: > I could'nt add above IF condition with != nil , since the compiler complains like
> 
> ```
> pkg/driver/controller_server.go:870:60: cannot convert nil to type model.SecondaryBackendDetails
> ```
> 
> I have confirmed the code changes with or without `SecondaryBackendDetails` populated via the CSP

ah, sorry i thought its a pointer. sounds good then.

pkg/driver/constants.go
```diff
@@ -35,11 +35,14 @@ const (
...
-	targetNamesKey        = "targetNames"
-	targetScopeKey        = "targetScope"
-	lunIDKey              = "lunId"
+	serialNumberKey   = "serialNumber"
+	accessProtocolKey = "accessProtocol"
+	targetNamesKey    = "targetNames"
+	targetScopeKey    = "targetScope"
+	lunIDKey          = "lunId"
+
+	secondaryArrayDetailsKey = "secondArrayDetailsKey"
```
rgcostea, line 44: "secondaryArrayDetailsKey"

wdurairaj: Fixed

shivamerla:
@wdurairaj common host libs is merged now. Can you update vendor directory here with "go mod edit --require github.com/hpe-storage/common-host-libs@master" and "go mod vendor"

wdurairaj:
> @wdurairaj common host libs is merged now. Can you update vendor directory here with "go mod edit --require github.com/hpe-storage/common-host-libs@master" and "go mod vendor"

@shivamerla , i have done the vendor updates.

raunakkumar:
@wdurairaj looks like there is a linter error with your last change, could you please fix it
```
pkg/driver/controller_server.go:877:10: `if` block ends with a `return` statement, so drop this `else` and outdent its block (golint)
		} else {
```

wdurairaj:
> @wdurairaj looks like there is a linter error with your last change, could you please fix it
> 
> ```
> pkg/driver/controller_server.go:877:10: `if` block ends with a `return` statement, so drop this `else` and outdent its block (golint)
> 		} else {
> ```

willl fix it .. looking into it

PR #190, opened 2020-07-01
implementing NodeGetVolumeStats RPC call

Baykonur:
Signed-off-by: Onur Bacak <obacak@epo.org>


rgcostea:
Hi @Baykonur .  Thanks for the PR.  We have been discussing internally about how best to implement the GET_VOLUME_STATS capability in our CSI driver.  The challenge is each filesystem may report these stats differently and the mount options affect how the stats are reported.  And none of this applies when raw block volumes are used.  So we will likely delegate the volume space reporting to the CSP representing the storage family (Nimble Storage, 3Par, Primera, etc.).

Baykonur:
Thanks @rgcostea. 
Is it correct then to say that the [CSP REST API](https://github.com/hpe-storage/container-storage-provider/blob/master/spec.md) will be used for this? I was actually looking there first for the concerns you pointed in your comment but there is only capacity information present (which might be OK for raw block volumes) at the moment.

Do you have any timeline for the API change and the implementation in the driver?

rgcostea:
Yes.  The CSP REST API will be used for this.  We will have an updated version of the spec that calls out the requirement to report used and free space along with capacity.  I can't really comment about the timeline but can you describe your use case.  Our product manager can take a look and then prioritize accordingly.  Thanks.

Baykonur:
What I implemented in this PR would be sufficient for us as we have Prometheus scraping `kubelet_volume_stats_*`.

I am closing this PR as this is in your road map, however it would be good to have a timeline.

PR #192, opened 2020-07-02
chap: retrieve chap credentials from configMap

raunakkumar:
- current implementation read chap user statically from iscsid.conf
This was not scaleable for node scale up/down scenarios
- read chap credentials from configMap and update the hpenodeinfo

### DIFF
--- a/cmd/csi-driver/csi-driver.go
+++ b/cmd/csi-driver/csi-driver.go
@@ -123,13 +123,6 @@ func csiCliHandler(cmd *cobra.Command) error {
 			return fmt.Errorf("Unable to configure iscsid service, err %v", err.Error())
 		}
 
-		// check if chapSecret is set in chap.json
-		driver.ChapUsername, driver.ChapPassword, err = tunelinux.GetIscsiChapSettings()
-		if err != nil {
-			// log error but proceed ahead
-			log.Warnf("unable to retrieve chap secrets ", err.Error())
-		}
-
 		// configure multipath
 		err = tunelinux.ConfigureMultipath()
 		if err != nil {
--- a/cmd/csi-driver/csi-driver.go
+++ b/cmd/csi-driver/csi-driver.go
@@ -123,6 +123,13 @@ func csiCliHandler(cmd *cobra.Command) error {
 			return fmt.Errorf("Unable to configure iscsid service, err %v", err.Error())
 		}
 
+		// check if chapSecret is set in chap.json
+		driver.ChapUsername, driver.ChapPassword, err = tunelinux.GetIscsiChapSettings()
+		if err != nil {
+			// log error but proceed ahead
+			log.Warnf("unable to retrieve chap secrets ", err.Error())
+		}
+
 		// configure multipath
 		err = tunelinux.ConfigureMultipath()
 		if err != nil {
--- a/pkg/driver/node_server.go
+++ b/pkg/driver/node_server.go
@@ -33,6 +33,10 @@ var (
 	unstageLock            sync.Mutex
 	ephemeralPublishLock   sync.Mutex
 	ephemeralUnpublishLock sync.Mutex
+	// ChapUsername username to be read from hpe-linux-config configMap
+	ChapUsername = ""
+	// ChapPassword to be read from configMap if present
+	ChapPassword = ""
 )
 var isWatcherEnabled bool = false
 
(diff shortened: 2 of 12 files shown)


cmd/csi-driver/csi-driver.go
```diff
@@ -123,6 +123,13 @@ func csiCliHandler(cmd *cobra.Command) error {
 			return fmt.Errorf("Unable to configure iscsid service, err %v", err.Error())
 		}
 
+		// check if chapSecret is set in chap.json
+		driver.ChapUsername, driver.ChapPassword, err = tunelinux.GetIscsiChapSettings()
```
shivamerla, line 127: i am thinking if we need to add CHAP login in tunelinux at all. Since it has more of set/get recommendations pattern, chap doesn't fit into that as there is no recommendation for this. Can we directly fetch these from ENV params with node plugin and populate hpenodeinfos with that?

pkg/driver/node_server.go
```diff
@@ -448,11 +452,15 @@ func (driver *Driver) setupDevice(publishContext map[string]string) (*model.Devi
 		ConnectionMode: defaultConnectionMode,
 	}
 	if publishContext[accessProtocolKey] == iscsi {
-		chapInfo := &model.ChapInfo{
```
shivamerla, line 451: why to override this here? if CHAP is enabled, then can we make sure its always added in publish context?

pkg/driver/node_server.go
```diff
@@ -1852,6 +1860,13 @@ func (driver *Driver) nodeGetInfo() (string, error) {
 				if initiator.Chap != nil {
 					chapUsername = initiator.Chap.Name
 					chapPassword = initiator.Chap.Password
+				} else {
```
shivamerla, line 1863: I think this causes a mismatch if something is already set in iscsid.conf and provided values with csi-driver are different. With this version, can we ignore iscsid.conf and always use configured values during controller publish and node stage?

cmd/csi-driver/csi-driver.go
raunakkumar: are you suggesting we don't need the additional parameters configMap as well? I am fine with that approach but we already have quite a few env variables to overload the csi driver. 
Additionally this fits well with the iscsid parameters we deal with.
@rgcostea what do you recommend?

pkg/driver/node_server.go
raunakkumar: passwords are not returned from the apis on the array (e.g chap_users), so CSP never returns it, hence the override is needed for password update, username remains the same.
i can modify this to not override if the key doesn't match for other CSPs which return the password.

pkg/driver/node_server.go
raunakkumar: i had kept this for backward compatibility but agree iscsid.conf can be ignored to keep it simple.

raunakkumar:
update: made the changes to comply with environment variables, eliminating iscsid.conf reference, not involving tunelinux and setting the password after controller publish response. So either way works well. Will update the PR once configMap vs env variable is finalized.

cmd/csi-driver/csi-driver.go
rgcostea: CHAP settings are a property of iSCSI configurations.  Can you describe what the purpose of TuneLinux is exactly?  Why aren't we using a ConfigMap in that case?  Historical reasons?  What are the pros and cons of each approach?

pkg/driver/node_server.go
```diff
@@ -448,11 +452,15 @@ func (driver *Driver) setupDevice(publishContext map[string]string) (*model.Devi
...
-			Name:     publishContext[chapUsernameKey],
-			Password: publishContext[chapPasswordKey],
+		// ChapPassword may not be returned from publishContext
+		// Validate the username from configMap and compare it from that returned from publishContext, and use the password from configMap
+		if publishContext[chapUsernameKey] == ChapUsername {
+			chapInfo := &model.ChapInfo{
+				Name:     ChapUsername,
+				Password: ChapPassword,
+			}
+			volume.Chap = chapInfo
```
rgcostea, line 462: Why does this need to be in the publish context?  I don't think we need to store these credentials on a k8s object do we?  We pull them at ControllerPublishVolume time to give to the CSP for configuration and again at NodeStageVolume time when the device actually needs to be created.

pkg/driver/node_server.go
rgcostea: We aren't supporting different CHAP users from different nodes are we?

pkg/driver/node_server.go
shivamerla: yes, this would be perfect since we have these values with both controller/node plugin already.

raunakkumar: i am changing this in CSP to return the credentials from host object so that there is no change in the csi driver (controller/node)

pkg/driver/node_server.go
rgcostea: 👍

PR #195, opened 2020-07-14
csi-driver: deprecate allowOverride key in favor of allowMutations

shivamerla:
* Problem:
  * with introduction of volume mutator, allowMutations key was introduced to let
  * SC specify which properties can be mutated by PVC. Since allowOverrrides serves
  * same purpose during volume creation it will be confusing if we introduce two property
  * names and allowed parameters list might not match.
* Implementation:
  * Deprecate allowOverrides in favor of allowMutations key and maintain backward
  * compatibility for existing storage classes with allowOverrrides.
* Testing: tested with duplicate and just allowOverrides or allowMutations.
* Review: gcostea, rkumar

### DIFF
--- a/pkg/flavor/kubernetes/flavor.go
+++ b/pkg/flavor/kubernetes/flavor.go
@@ -35,6 +35,7 @@ import (
 
 const (
 	allowOverrides       = "allowOverrides"
+	allowMutations       = "allowMutations"
 	nodeUnreachableTaint = "node.kubernetes.io/unreachable"
 	nodeLostCondition    = "NodeLost"
 )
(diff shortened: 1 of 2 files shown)


pkg/flavor/kubernetes/flavor.go
```diff
@@ -377,28 +381,40 @@ func (flavor *Flavor) getClaimFromClaimName(name string) (*v1.PersistentVolumeCl
...
+	if _, ok := optionsMap[allowMutations]; ok {
+		if overrideKey != "" {
+			return nil, fmt.Errorf("Duplicate override keys are specified in StorageClass. Please use only '%s' key to specify parameters allowed to be overridden by the PVC", allowMutations)
+		}
+		overrideKey = allowMutations
+	}
+
+	var overrideParams []string
+	if val, ok := optionsMap[overrideKey]; ok {
+		log.Infof("allowed overrides in sc are %s", val)
```
raunakkumar, line 402: nit: storage class

shivamerla:
@raunakkumar closing this out as further discussion needed to avoid confusion.

PR #199, opened 2020-07-22
Fix ListSnapshots

rgcostea:
- Updated CSI spec reference to v1.3.0
- Added unimplemented function for ControllerGetVolume
- Referenced Secrets and SnapshotID when provided to the ListSnapshotsRequest


pkg/driver/controller_server.go
```diff
@@ -1350,28 +1369,43 @@ func (driver *Driver) ListSnapshots(ctx context.Context, request *csi.ListSnapsh
...
+			}, nil
+		}
+
+		for _, storageProvider := range driver.storageProviders {
+			snapshots, err := getSnapshotsForStorageProvider(ctx, request, storageProvider)
+			if err != nil {
+				return nil, err
+			}
+
+			if len(snapshots) > 0 {
```
shivamerla, line 1404: Since it seems like we always have either snapshotID or sourceVolumeID, can we break the loop here once we get snapshots from one provider? If both of these can be nil, then do we still need to fetch all snapshots of all volumes and all providers in that case?

rgcostea: We sure can.  Thanks for that idea.

pkg/driver/controller_server.go
```diff
@@ -1350,28 +1369,45 @@ func (driver *Driver) ListSnapshots(ctx context.Context, request *csi.ListSnapsh
...
+		if len(driver.storageProviders) == 0 {
+			log.Info("No storage providers are known to the CSI driver yet.")
+			return &csi.ListSnapshotsResponse{
+				Entries: []*csi.ListSnapshotsResponse_Entry{},
+			}, nil
+		}
+
+		for _, storageProvider := range driver.storageProviders {
+			snapshots, err := getSnapshotsForStorageProvider(ctx, request, storageProvider)
+			if err != nil {
```
shivamerla, line 1400: we should not error out here then as it might be wrong storageprovide for that volume/snapshotid.

rgcostea: I don't think we should ignore all errors here.  If it is the wrong storage provider, we will get a 404 error and err will not be nil.  If there is an auth issue, we get a 401 and login again.

If it is the right storage provider but there's an error, we need to log it and return.

shivamerla: yes, this will be good. we seem to get error codes in string, so can lookup these.

PR #213, opened 2020-10-06
Update the conform script to include Amazon for installing iscsi

ckallianpur:
Signed-off-by: chaitra <chaitra.kallianpur@hpe.com>
On Amazon EKS, the default EC2 node is set up using Amazon Linux, so it fails this install step. It can be fixed by adding Amazon Linux to the "Redhat | CentOS" check, since it is based on those and provides the necessary packages for iscsi.


raunakkumar (approved):

suneeth51 (approved):
lgtm..

PR #216, opened 2020-10-14
csi: deprecate protectionTemplate from volume create options

raunakkumar:

### DIFF
--- a/pkg/driver/controller_server.go
+++ b/pkg/driver/controller_server.go
@@ -341,6 +341,11 @@ func (driver *Driver) createVolume(
 		return nil, status.Error(codes.Internal, errStr)
 	}
 
+	if val, ok := createParameters[protectionTemplateKey]; ok {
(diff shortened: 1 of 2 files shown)


pkg/driver/controller_server.go
```diff
@@ -341,6 +341,11 @@ func (driver *Driver) createVolume(
 		return nil, status.Error(codes.Internal, errStr)
 	}
 
+	if val, ok := createParameters[protectionTemplateKey]; ok {
+		log.Infof("protectionTemplate is deprecated for volume create, removing %s=%s from the request", protectionTemplateKey, val)
```
rgcostea, line 345: This is more than deprecating :)

This is removal.

rgcostea (approved):

pkg/driver/controller_server.go
datamattsson: Can we change the verbiage to mention that volumeGroups are used instead and the parameter has "moved"?

raunakkumar: sure. will update

PR #221, opened 2020-10-29
Con 1072 - Fix code to resolve the driver sanity test failures

ckallianpur:


pkg/driver/controller_server.go
```diff
@@ -694,6 +694,15 @@ func (driver *Driver) ControllerPublishVolume(ctx context.Context, request *csi.
 		return nil, status.Error(codes.InvalidArgument, "Invalid volume capability specified for ControllerPublishVolume")
 	}
 
+	roVal, err := strconv.ParseBool(request.VolumeContext[readOnlyKey])
```
raunakkumar, line 697: nit: we can rename it to `readOnlyValue` just like the key

pkg/driver/node_server.go
```diff
@@ -1656,6 +1656,11 @@ func (driver *Driver) NodeGetVolumeStats(ctx context.Context, in *csi.NodeGetVol
 	log.Trace(">>>>> NodeGetVolumeStats")
 	defer log.Trace("<<<<< NodeGetVolumeStats")
 
+	volumeID := in.GetVolumeId()
+	if volumeID == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "volume id %s is empty", volumeID)
```
raunakkumar, line 1661: we didn't need the volumeID in the message if its empty

pkg/driver/controller_server.go
ckallianpur: upadated

pkg/driver/node_server.go
ckallianpur: fixed

raunakkumar (approved):

pkg/driver/node_server.go
```diff
@@ -1693,18 +1698,18 @@ func (driver *Driver) NodeGetVolumeStats(ctx context.Context, in *csi.NodeGetVol
...
-	volume, err := storageProvider.GetVolume(in.VolumeId)
+	volume, err := storageProvider.GetVolume(volumeID)
 	if err != nil {
 		log.Error("err: ", err.Error())
 		return nil, status.Errorf(codes.Internal, "Error while retrieving volume with id %s from the backend, err: %s",
-			in.VolumeId, err.Error())
+			volumeID, err.Error())
 	}
 	if volume == nil {
 		log.Infof("Volume with ID %s not found on the backend, return ",
```
suneeth51, line 1711: nit: log.Errorf("Volume with ID %s not found on the backend")

pkg/driver/node_server.go
```diff
@@ -1693,18 +1698,18 @@ func (driver *Driver) NodeGetVolumeStats(ctx context.Context, in *csi.NodeGetVol
...
-			in.VolumeId)
+			volumeID)
 		return nil, nil
```
suneeth51, line 1713: Shouldn't we return `codes.NotFound` ?

pkg/driver/controller_server.go
```diff
@@ -694,6 +694,15 @@ func (driver *Driver) ControllerPublishVolume(ctx context.Context, request *csi.
 		return nil, status.Error(codes.InvalidArgument, "Invalid volume capability specified for ControllerPublishVolume")
 	}
 
+	readOnlyValue, err := strconv.ParseBool(request.VolumeContext[readOnlyKey])
+	if err != nil {
+		log.Tracef("Failed to retrieve readonly key from volumeContext : %v", err.Error())
```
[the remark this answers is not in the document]
ckallianpur: I think this is wrong. If the volume has readonly and if its value is different from the value passed via request, we need to log this error. I didnt find ReadOnly property returned from csp.

suneeth51 (approved):
lgtm..

PR #227, opened 2020-12-02
CON-1600: Create target_path during publish if it doesnt exist

ckallianpur:
target_path neednt exist in request params in NodePublishVolume from kubernetes 1.20 onwards. Create it if it doesnt exist


raunakkumar (approved):
lgtm. We can try this patch with 1.20 which releases tomorrow.

PR #241, opened 2021-02-03
CON-1694: Update nodeinfo with chap credentials if provided by publis…

ckallianpur:
…hcontext

### DIFF
--- a/pkg/driver/controller_server.go
+++ b/pkg/driver/controller_server.go
@@ -926,13 +926,10 @@ func (driver *Driver) controllerPublishVolume(
 		publishContext[fsCreateOptionsKey] = volumeContext[fsCreateOptionsKey]
 	}
 
-	sensitive := map[string]string{}
+	sensitive := []string{}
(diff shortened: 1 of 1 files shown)


pkg/driver/controller_server.go
```diff
@@ -913,8 +925,15 @@ func (driver *Driver) controllerPublishVolume(
 		publishContext[fsModeKey] = volumeContext[fsModeKey]
 		publishContext[fsCreateOptionsKey] = volumeContext[fsCreateOptionsKey]
 	}
+
+	sensitive := []string{}
+	for key, value := range publishContext {
+		sensitive = append(sensitive, key)
```
raunakkumar, line 931: nit: sensitive = append(sensitive, key, value)

raunakkumar (approved):
lgtm

pkg/driver/controller_server.go
ckallianpur: done

PR #246, opened 2021-02-22
Volume encryption support

imran-ansari:
Volume encryption support is a host-based encryption using LUKS tool called cryptsetup. During node-stage flow, cryptsetup tool is used on the discovered device to enable encryption. For details about the approach please refer to https://rndwiki.corp.hpecorp.net/confluence/display/storageecosystem/Volume+encryption+support+in+CSI+Driver

@rgcostea , @raunakkumar , @wdurairaj , @sneharai4 - request you to review the changes

### DIFF
--- a/cmd/csi-driver/Dockerfile
+++ b/cmd/csi-driver/Dockerfile
@@ -1,12 +1,9 @@
 FROM registry.access.redhat.com/ubi7/ubi-minimal:7.9-256
 
-RUN microdnf install -y yum
 RUN microdnf update -y && rm -rf /var/cache/yum
-RUN curl https://rpmfind.net/linux/centos/7.9.2009/os/x86_64/Packages/cryptsetup-2.0.3-6.el7.x86_64.rpm -o cryptsetup-2.0.3-6.el7.x86_64.rpm
-RUN yum install -y cryptsetup-2.0.3-6.el7.x86_64.rpm
-RUN rm -f cryptsetup-2.0.3-6.el7.x86_64.rpm
 RUN microdnf install -y systemd
-RUN microdnf remove -y yum
+ADD CentOS-Base.repo /etc/yum.repos.d/
+RUN microdnf install -y cryptsetup
 
 LABEL name="HPE CSI Driver for Kubernetes" \
       maintainer="HPE Storage" \
--- a/cmd/csi-driver/Dockerfile
+++ b/cmd/csi-driver/Dockerfile
@@ -1,6 +1,10 @@
 FROM registry.access.redhat.com/ubi7/ubi-minimal:7.9-256
 
+RUN microdnf install -y yum
 RUN microdnf update -y && rm -rf /var/cache/yum
+RUN curl https://rpmfind.net/linux/centos/7.9.2009/os/x86_64/Packages/cryptsetup-2.0.3-6.el7.x86_64.rpm -o cryptsetup-2.0.3-6.el7.x86_64.rpm
+RUN yum install -y cryptsetup-2.0.3-6.el7.x86_64.rpm
+RUN rm -f cryptsetup-2.0.3-6.el7.x86_64.rpm
 RUN microdnf install -y systemd
 
 LABEL name="HPE CSI Driver for Kubernetes" \
--- a/cmd/csi-driver/Dockerfile
+++ b/cmd/csi-driver/Dockerfile
@@ -6,6 +6,7 @@ RUN curl https://rpmfind.net/linux/centos/7.9.2009/os/x86_64/Packages/cryptsetup
 RUN yum install -y cryptsetup-2.0.3-6.el7.x86_64.rpm
 RUN rm -f cryptsetup-2.0.3-6.el7.x86_64.rpm
 RUN microdnf install -y systemd
+RUN microdnf remove -y yum
 
 LABEL name="HPE CSI Driver for Kubernetes" \
       maintainer="HPE Storage" \
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -95,7 +95,6 @@ const (
 	// secretName attribute to be used to specify secret name for ephemeral inline volume
 	secretNameKey = "inline-volume-secret-name"
 
-	volEncrypted = "volumeEncrypted"
 	volEncryptionSecretNameKey = "encryptionSecretName"
 	volEncryptionSecretNamespaceKey = "encryptionSecretNamespace"
 	volEncryptionKey = "volumeEncryptionKey"
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -95,6 +95,11 @@ const (
 	// secretName attribute to be used to specify secret name for ephemeral inline volume
 	secretNameKey = "inline-volume-secret-name"
 
+	volEncrypted = "volumeEncrypted"
+	volEncryptionSecretNameKey = "encryptionSecretName"
+	volEncryptionSecretNamespaceKey = "encryptionSecretNamespace"
+	volEncryptionKey = "volumeEncryptionKey"
+
 	// POD attributes propogated to the CSI
 	csiEphemeralPodName      = "csi.storage.k8s.io/pod.name"
 	csiEphemeralPodNamespace = "csi.storage.k8s.io/pod.namespace"
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -95,9 +95,10 @@ const (
 	// secretName attribute to be used to specify secret name for ephemeral inline volume
 	secretNameKey = "inline-volume-secret-name"
 
-	volEncryptionSecretNameKey = "encryptionSecretName"
-	volEncryptionSecretNamespaceKey = "encryptionSecretNamespace"
-	volEncryptionKey = "volumeEncryptionKey"
+	hostEncryptionKey = "hostEncryption"
+	hostEncryptionSecretNameKey = "hostEncryptionSecretName"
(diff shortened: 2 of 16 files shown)


cmd/csi-driver/Dockerfile
```diff
@@ -1,7 +1,12 @@
 FROM registry.access.redhat.com/ubi7/ubi-minimal:7.9-256
 
+RUN microdnf install -y yum
 RUN microdnf update -y && rm -rf /var/cache/yum
+RUN curl https://rpmfind.net/linux/centos/7.9.2009/os/x86_64/Packages/cryptsetup-2.0.3-6.el7.x86_64.rpm -o cryptsetup-2.0.3-6.el7.x86_64.rpm
```
wdurairaj, line 5: OSRB approval may be required here for inclusion of this RPM in UBI image

cmd/csi-driver/Dockerfile
```diff
@@ -1,7 +1,12 @@
...
+RUN yum install -y cryptsetup-2.0.3-6.el7.x86_64.rpm
+RUN rm -f cryptsetup-2.0.3-6.el7.x86_64.rpm
 RUN microdnf install -y systemd
+RUN microdnf remove -y yum
```
datamattsson, line 9: This is very ugly. Please add a file named `CentOS-Base.repo` to this repo with the following content:
```
[base]
name=CentOS-7 - Base
mirrorlist=http://mirrorlist.centos.org/?release=7&arch=$basearch&repo=os&infra=$infra
gpgcheck=0
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7

[updates]
name=CentOS-7 - Updates
mirrorlist=http://mirrorlist.centos.org/?release=7&arch=$basearch&repo=updates&infra=$infra
gpgcheck=0
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
```

Then replace line 3-9 with:
```
ADD CentOS-Base.repo /etc/yum.repos.d/
```
Line 4 (currently 8) should then read:
```
RUN microdnf install -y systemd cryptsetup
```
This is the pattern we've established with other external dependencies (such as NFS).

datamattsson (changes requested):

pkg/driver/constants.go
```diff
@@ -95,6 +95,10 @@ const (
 	// secretName attribute to be used to specify secret name for ephemeral inline volume
 	secretNameKey = "inline-volume-secret-name"
 
+	volEncryptionSecretNameKey = "encryptionSecretName"
+	volEncryptionSecretNamespaceKey = "encryptionSecretNamespace"
+	volEncryptionKey = "volumeEncryptionKey"
+
```
datamattsson, line 101: After some debate with the PM/architects, these parameters needs to be changed into:
```
hostEncryptionSecretName
hostEncryptionSecretNamespace
```
The `Secret` key needs to be named:
```
hostEncryptionKey
```
We also need to re-introduce the boolean. This will be used for easy toggling of the host encryption feature from a PVC annotation (override during provisioning only, not mutable).

The boolean should be named:
```
hostEncryption
```

pkg/driver/node_server.go
```diff
@@ -256,6 +256,27 @@ func (driver *Driver) nodeStageVolume(
 		return nil // volume already staged, do nothing and return here
 	}
 
+	if volumeContext[volEncryptionSecretNameKey] != "" {
```
datamattsson, line 259: Change to boolean as per above request.

datamattsson (changes requested):

cmd/csi-driver/Dockerfile
wdurairaj: @datamattsson , we are not adding any CentOS repository package for NFS provisioner , rather we pull the nfs provisioner image as a container and we run ganeshd as a container in pod

https://github.com/hpe-storage/csi-driver/blob/be80c2e6252d89f8d463579ab1338390e7ad7296/pkg/flavor/kubernetes/nfs.go#L27

We will explore the option of adding this centos repository as suggested by you.

datamattsson: I was merely referring to the pattern of adding packages from outside of what is available in the UBI repo.

imran-ansari: Thanks @datamattsson - nice input! Incorporated it locally and did some testing. Once all the review comments are addressed, I will push the changes.

pkg/driver/constants.go
imran-ansari: Will rename the parameter name to have "host" prefix rather than "volume" prefix. 

As for toggling of encryption feature is concerned, once a PVC is created, be it encrypted or unencrypted, we don't support switching it to the other type.

Besides, I would like to highlight the following points with respect to the boolean hostEncryption:
1. In order to use this feature, user would need to specify an additional parameter in storage class
2. hostEncryption boolean is redundant when the presence of hostSecretName and hostSecretNamespace suffices.
3. Unnecessary validations would be required when user would be specifying one parameter while skipping the other.
4. Configurations like hostEncryption set to false while specifying hostSecrentName and hostSecretNamespace wouldn't make any sense.

In view of the above, I would suggest that we don't introduce one more parameter.

pkg/driver/node_server.go
imran-ansari: This will still be needed to validate the user supplied value.

pkg/driver/constants.go
datamattsson: The user experience we're trying to cater for is targeted towards two personas. The Kubernetes admin and Kubernetes user, normally locked down to certain namespaces. Kubernetes users do not have the ability to create a StorageClass as it's a cluster wide object.

I understand your concern that that the `hostEncryption` may seem redundant but it makes the user experience a bit broken if we omit it entirely. We want to cater to as many use cases as possible. So, let's discuss those as I think there's a middle road here.

- ### As a Kubernetes admin, I want to enforce `hostEncryption` for all PVCs created from my StorageClass using my host key.
  
  **StorageClass**:
  ```
  parameters:
    hostEncryptionSecretName: my-super-secret
    hostEncryptionSecretNamespace: hpe-storage
  ```
  In this case, `hostEncryption` boolean will be redundant, setting `hostEncryptionSecretName` and `hostEncryptionSecretNamespace` will automatically qualify `hostEncryption` to "true".

- ### As a Kubernetes admin, I want to allow users to selectively encrypt PVCs created from my StorageClass using my host key.

  **StorageClass**:
  ```
  parameters:
    allowOverrides: hostEncryption
    hostEncryption: "false"
    hostEncryptionSecretName: my-super-secret
    hostEncryptionSecretNamespace: hpe-storage
  ```

  **PVC**:
  ```
  metadata:
    annotations:
       csi.hpe.com/hostEncryption: "true"
  ```
  Here the `hostEncryption` boolean is critical, for two reasons. The Kubernetes user has no clue where to find the host key secret and that needs to be provided by the StorageClass. Second, without `hostEncryption`, the user would have to set the secret and secretname to empty strings and it would then allow users to specify their own host key, which is not desired, since the admin would have to allow both parameters to be overridden.
  
  The admin may use this pattern to allow users to opt-out from host encryption by using `hostEncryption` set to "true" and the annotation set to "false".

- ### As a Kubernetes admin, I want to allow users to selectively encrypt PVCs created from my StorageClass using their own host keys.

  **StorageClass**:
  ```
  parameters:
    allowOverrides: hostEncryptionSecretName,hostEncryptionSecretNamespace
  ```

  **PVC**:
  ```
  metadata:
    annotations:
       csi.hpe.com/hostEncryptionSecretName: my-secret
       csi.hpe.com/hostEncryptionSecretNamespace: my-namespace
  ```
  In this case, `hostEncryption` wouldn't be necessary as both parameters will be overridden with the users annotations.

imran-ansari: @datamattsson - thank you for the detailed use cases. There is a catch there though which we would need to document. For an existing PVC having user data on it, if any of the mentioned use cases is executed by the user, the data on the PVC would be wiped out.

datamattsson: @imran-ansari please elaborate. The override annotations are only available to the user during provisioning of a new PVC. All parameters related to host encryption should be immutable once the PVC is provisioned. Not sure if the volume-mutator sidecar or the CSI driver has a blocklist. Anyhow, we can document it either way.

imran-ansari: @datamattsson  - Please ignore my previous comment. Since `allowOverride` is for new volume creation, my above observation doesn't hold good. I'll go ahead with the flag addition and push the changes again.

pkg/driver/node_server.go
```diff
@@ -256,6 +256,27 @@ func (driver *Driver) nodeStageVolume(
...
+			return err // specified secret for encryption key doesn't exist
+		}
+
+		ctxt := make(map[string]string)
+		for k,v := range publishContext {
+			ctxt[k] = v
+		}
+		ctxt[volEncryptionKey] = encPassphrase
+		publishContext = ctxt
+		log.Tracef("Modified publish context: %v", publishContext)
```
sneharai4, line 277: Are we printing the encPassphrase here as part of publishContext?

imran-ansari: This was for debugging purpose. Will be removing it.

datamattsson:
Thanks for addressing our UX concerns, LGTM.

raunakkumar (approved):
lgtm

raunakkumar (approved):

raunakkumar (approved):

PR #253, opened 2021-03-30
csi: NodeGetVolumeStats fetch stats from Statfs

raunakkumar:
- with large scale setups, polling csp every minute to fetch stats causes bottleneck to the CSPs
- the stats for bytes free and used can be fetched from Statfs

Test Results: 
```
time="2021-03-30T16:05:21Z" level=info msg="GRPC response: {\"usage\":[{\"available\":10598957056,\"total\":10737418240,\"unit\":1,\"used\":138461184},{\"available\":5242876,\"total\":5242880,\"unit\":2,\"used\":4}]}" file="utils.go:75"
```

### DIFF
--- a/pkg/driver/node_server.go
+++ b/pkg/driver/node_server.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -21,7 +22,6 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/api/resource"
-	vol "k8s.io/kubernetes/pkg/volume"
 	"k8s.io/utils/mount"
 
 	log "github.com/hpe-storage/common-host-libs/logger"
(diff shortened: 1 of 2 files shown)


pkg/driver/node_server.go
```diff
@@ -1801,63 +1801,25 @@ func (driver *Driver) NodeGetVolumeStats(ctx context.Context, in *csi.NodeGetVol
...
-			volumeID, err.Error())
-	}
-	if volume == nil {
-		log.Errorf("Volume with ID %s not found on the backend, return ", volumeID)
-		return nil, status.Errorf(codes.NotFound, "Volume with ID %s not found on the backend, return ",
-			volumeID)
-	}
-	hpeMetricsProvider := vol.NewMetricsStatFS(in.VolumePath)
-	metrics, err := hpeMetricsProvider.GetMetrics()
+	var statfs syscall.Statfs_t
```
suneeth51, line 1805: Ideally, this functionality should be implemented in the `common-host-libs/linux package` and call that method here. In the future, let's plan to move it there.

suneeth51 (approved):

raunakkumar:
As discussed, we will take this feature for next release.  Marking it as closed for 1.5

datamattsson:
Initially we wanted a solution that works both for `volumeMode: Block` and `volumeMode: Filesystem`. Not sure statfs will give us that.

raunakkumar:
Yes agree. I did some research offline and other vendors are using kubernetes/pkg/volume library (https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/metrics_statfs.go#L42) to get stats. For block volume the block size bytes can be retrieved like the aws ebs csi driver below
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/driver/node.go#L408-L467

PR #254, opened 2021-03-30
 Removed chap credentials from the PublishContext

suneeth51:
- Removed chap credentials from the PublishContext
- For CV-CSP, the chap credentials are fetched from the Volume response.
- For Nimble-CSP and 3PAR/Primera-CSP, the chap credentials are retrieved from the `hpeNodeInfos` CRD


pkg/driver/node_server.go
```diff
@@ -498,21 +534,15 @@ func (driver *Driver) setupDevice(publishContext map[string]string) (*model.Devi
 		DiscoveryIPs:          discoveryIps,
 		ConnectionMode:        defaultConnectionMode,
 		SecondaryArrayDetails: publishContext[secondaryArrayDetailsKey],
-		EncryptionKey: 		   publishContext[hostEncryptionPassphraseKey],
-	}
-	if publishContext[accessProtocolKey] == iscsi {
```
raunakkumar, line 503: we should still set CHAP only if protocol == iscsi

raunakkumar (approved):
lgtm

pkg/driver/controller_server.go
```diff
@@ -807,11 +806,6 @@ func (driver *Driver) controllerPublishVolume(
 		return nil, status.Error(codes.NotFound, err.Error())
 	}
 
-	if node.ChapUser != "" && node.ChapPassword != "" {
-		decodedChapPassword, _ := b64.StdEncoding.DecodeString(node.ChapPassword)
```
raunakkumar, line 811: Hey @suneeth51 I was just reviewing this with @sneharai4 and we realized that we need to return unencrypted password to CSP.

raunakkumar (changes requested):

pkg/driver/controller_server.go
suneeth51: Sure, I will fix that.

raunakkumar (approved):

sneharai4 (approved):

PR #265, opened 2021-05-05
csi: support for alternate kubelet root directory

raunakkumar:


suneeth51 (changes requested):
As discussed, you can move the env variable read within NewDriver() logic and avoid passing it down as function param.

suneeth51 (approved):
lgtm..

PR #268, opened 2021-05-07
Added support to call the NodeGetVolumeStats as per the env variable set 

bhagyashree-sarawate:
Bydefault NodeGetVolumeStats will be called 
Setting the env variable to enable/disable the call 
@sneharai4 @raunakkumar please review

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+	"strconv"
 
 	"github.com/Scalingo/go-etcd-lock/lock"
 	"github.com/container-storage-interface/spec/lib/go/csi"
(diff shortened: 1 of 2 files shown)


pkg/driver/driver.go
```diff
@@ -243,9 +244,14 @@ func (driver *Driver) AddControllerServiceCapabilities(capabilities []csi.Contro
 func (driver *Driver) AddNodeServiceCapabilities(capabilities []csi.NodeServiceCapability_RPC_Type) {
 	var nodeServiceCapabilities []*csi.NodeServiceCapability
 
+	disableNodeGetVolumeStats, _ := strconv.ParseBool(os.Getenv(disableNodeGetVolumeStatsKey))
```
raunakkumar, line 247: nit: please return the error in case user enters non boolean flag.

raunakkumar (approved):
lgtm

PR #275, opened 2021-05-25
update chap on csi-driver upgrade

raunakkumar:
Problem: we don't manage CRDs between releases with helm.
This leaves the hpenodeinfo crd as is between releases for chap updates
Implementation: update chap just like other hpenodeinfo params

### DIFF
--- a/pkg/flavor/kubernetes/flavor.go
+++ b/pkg/flavor/kubernetes/flavor.go
@@ -41,6 +41,8 @@ const (
 	nodeLostCondition             = "NodeLost"
 	provisionerSecretNameKey      = "csi.storage.k8s.io/provisioner-secret-name"
 	provisionerSecretNamespaceKey = "csi.storage.k8s.io/provisioner-secret-namespace"
+	chapUserEnvKey                = "CHAP_USER"
+	chapPasswordEnvKey            = "CHAP_PASSWORD"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/flavor/kubernetes/flavor.go
```diff
@@ -281,6 +293,14 @@ func getWwpnsFromNode(node *model.Node) []string {
 	return wwpns
 }
 
+func getChapUserFromEnvironment() string {
+	return os.Getenv("CHAP_USER")
```
sneharai4, line 297: Can we use here, chapUserEnvKey const?

pkg/flavor/kubernetes/flavor.go
```diff
@@ -281,6 +293,14 @@ func getWwpnsFromNode(node *model.Node) []string {
...
+}
+
+func getChapPasswordFromEnvironment() string {
+	return os.Getenv("CHAP_PASSWORD")
```
sneharai4, line 301: Can we use here, chapPasswordEnvKey const?

pkg/flavor/kubernetes/flavor.go
raunakkumar: Importing constants creates a cyclic import dependency.

sneharai4: Can we update https://github.com/hpe-storage/csi-driver/blob/master/pkg/driver/node_server.go#L2059
https://github.com/hpe-storage/csi-driver/blob/master/pkg/driver/node_server.go#L2060 to call the new method?

sneharai4 (approved):

pkg/flavor/kubernetes/flavor.go
```diff
@@ -224,12 +234,24 @@ func (flavor *Flavor) LoadNodeInfo(node *model.Node) (string, error) {
...
 
+		chapUser := flavor.GetChapUserFromEnvironment()
+		if !reflect.DeepEqual(nodeInfo.Spec.ChapUser, chapUser) {
+			nodeInfo.Spec.ChapUser = chapUser
+			updateNodeRequired = true
+		}
+
+		chapPassword := flavor.GetChapPasswordFromEnvironment()
+		if !reflect.DeepEqual(nodeInfo.Spec.ChapPassword, chapPassword) {
+			nodeInfo.Spec.ChapPassword = chapPassword
```
sneharai4, line 245: Do we need to update chap with encoded value like we do here https://github.com/hpe-storage/csi-driver/blob/master/pkg/flavor/kubernetes/flavor.go#L250?

raunakkumar: Yes, thanks for catching this.

suneeth51 (approved):

PR #281, opened 2021-06-21
Added resolved issues for Primera 3PAR

bhagyashree-sarawate:
@pavansshanbhag @sneharai4 @Raunak Please review

Added the resolved fix from last release

### DIFF
--- a/release-notes/v2.0.0.md
+++ b/release-notes/v2.0.0.md
@@ -10,6 +10,16 @@
 
 The following table lists the resolved issues for HPE CSI Driver for Kubernetes v2.0.0.
 
+|ID|Component |Title|Resolution|
+|--|---------|-----|-----------|
+|BZ-134925|csi.k8s|Disable NodeGetVolumeStats capability|Set the parameter `disableNodeGetVolumeStats` to true at the time of deployment.|
+
+The following table lists the resolved issues for HPE Primera/3PAR CSP for HPE CSI Driver release 2.0.0.
+
+|ID|Component |Title|Resolution|
+|--|---------|-----|-----------|
+|BZ-134504|csi.k8s| More online FC ports caused huge time for mount| New parameter `fcPortsList` is added in Storageclass and will be applicable only when a new volume is created.|
+
 ## Known Issues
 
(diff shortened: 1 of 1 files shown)


release-notes/v2.0.0.md
```diff
@@ -10,6 +10,16 @@
...
 
+|ID|Component |Title|Resolution|
+|--|---------|-----|-----------|
+|BZ-134925|csi.k8s|Disable NodeGetVolumeStats capability|Set the parameter `disableNodeGetVolumeStats` to true at the time of deployment.|
+
+The following table lists the resolved issues for HPE Primera/3PAR CSP for HPE CSI Driver release 2.0.0.
+
+|ID|Component |Title|Resolution|
+|--|---------|-----|-----------|
+|BZ-134504|csi.k8s| More online FC ports caused huge time for mount| New parameter `fcPortsList` is added in Storageclass and will be applicable only when a new volume is created.|
```
pavansshanbhag, line 21: Consider rewording as below:
|BZ-134504|csi.k8s| Suboptimal attach/detach performance when all FC ports are auto-discovered and used for multipathing| For a higher end primera array, there could be several FC ports available for use with multipathing. By default the CSP auto-discovers all available FC ports and uses them for multipathing. This increases the time taken for attach/detach operations significantly. To address this, a new parameter `fcPortsList` is added in Storageclass. This will be applicable only for newer volumes that are created.|

release-notes/v2.0.0.md
```diff
@@ -10,6 +10,16 @@
 
 The following table lists the resolved issues for HPE CSI Driver for Kubernetes v2.0.0.
```
pavansshanbhag, line 11: Can we combine these two know issues into a single table ?

pavansshanbhag (changes requested):
Please see comments provided inline

release-notes/v2.0.0.md
bhagyashree-sarawate: One is a CSI driver specific and other is csp specific , so made two different tables as per the last release notes ( We have addressed in two different places)

raunakkumar (approved):

release-notes/v2.0.0.md
bhagyashree-sarawate: done

pavansshanbhag (approved):

PR #293, opened 2021-12-01
Switching to UBI-8

datamattsson:


cmd/csi-driver/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi7/ubi-minimal:7.9
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.5-204
```
raunakkumar, line 1: lets keep this as `8.5` as that would point to latest 8.5 tag instead on going to minor version

raunakkumar (approved):

PR #296, opened 2022-01-10
Release 2.1

maddiegos:
release notes as per doc task 
https://nimblejira.nimblestorage.com/browse/CON-2576


release-notes/v2.1.0.md
```diff
@@ -0,0 +1,36 @@
...
+
+|ID|Component |Title|Description|Workaround|
+|--|---------|-----|-----------|----------|
+
+
+The following table lists the known issues for HPE Alletra 9000 and Primera/3PAR CSP for HPE CSI Driver release 2.1.0
+
+|ID|Component |Title|Description|Workaround|
+|--|---------|-----|-----------|----------|
+|CON-2330|csi.k8s|Discrepancy in volume size between array and master. | When volume size is not provided in multiple of 256 then array round it off to nearby 256 multiple value but on the node level it's size get not reflected. |None|
```
pavansshanbhag, line 34: Description: When volume size is not provided in multiples of 256 Mib, the array provisions the volume with a size that is rounded off to the nearest 256 multiple, however the k8s shows the original (unrounded) volume size, hence the discrepancy.

release-notes/v2.1.0.md
```diff
@@ -0,0 +1,36 @@
...
+|CON-2536|csi.k8s| RCG Greater than 23 characters generating errors. | When user provides RCG name more than 23 characters, array does not create rcg as it is against it's core behaviour |RCG name should always be less than 23 characcters. |
```
pavansshanbhag, line 35: Description: Volume mutation fails if the RCG name specified is invalid or longer than 22 characters. There is no feedback given to the end user about the invalid input, except in logs.
Workaround: User can mutate the PVC again providing a valid RCG name (should be less than or equal to 22 characters and should not contain special characters except ..TBD: please specify what it should contain)

release-notes/v2.1.0.md
```diff
@@ -0,0 +1,36 @@
...
+|CON-2463|csi.k8s|mutate pvc with oneRgcPerPvc: "!23false##", no further edits working and csp logs are looping continuously. | Not able to edit once incorrect values are provided. |As a solution we just needs to provide empty string for all the annotation related to remotecopy group mutation such as remoteCopyGroup, oneRcgPerPvc, replicationDevices then again user needs to enter the correct values . so this is a workaround for now. |
```
pavansshanbhag, line 36: Title : Mutating the oneRcgPerPvc attribute with invalid value (such as "!23false##") would render any subsequent mutations ineffective
Description: In addition the logs show continous activity. No feedback provided to kubernetes admin/user.
Workaround: Mutate the PVC with the annotation values of the attributes - remoteCopyGroup, oneRcgPerPvc, replicationDevices set to empty, and then mutate again to provide the accurate values for them.

pavansshanbhag (changes requested):

release-notes/v2.1.0.md
maddiegos: modified the same content

release-notes/v2.1.0.md
maddiegos: modified the same content except the special characters which is to be decided

release-notes/v2.1.0.md
maddiegos: modified the same content

pavansshanbhag (approved):

PR #298, opened 2022-02-02
Update CentosBase to AlmaLinux as Centos is EOL

raunakkumar:
https://www.centos.org/centos-linux-eol/

### DIFF
--- a/cmd/csi-driver/Dockerfile
+++ b/cmd/csi-driver/Dockerfile
@@ -7,7 +7,7 @@ RUN microdnf install -y systemd && microdnf install -y cryptsetup
 LABEL name="HPE CSI Driver for Kubernetes" \
       maintainer="HPE Storage" \
       vendor="HPE" \
-      version="2.1.1.0" \
+      version="2.1.1" \
       summary="HPE CSI Driver for Kubernetes" \
       description="The HPE CSI Driver for Kubernetes enables container orchestrators, such as Kubernetes and OpenShift, to manage the life-cycle of persistent storage." \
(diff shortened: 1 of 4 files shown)


cmd/csi-driver/Dockerfile
```diff
@@ -1,13 +1,13 @@
...
 RUN microdnf update -y && rm -rf /var/cache/yum
-ADD CentOS-Base.repo /etc/yum.repos.d/
+ADD AlmaLinux-Base.repo /etc/yum.repos.d/
 RUN microdnf install -y systemd && microdnf install -y cryptsetup
 
 LABEL name="HPE CSI Driver for Kubernetes" \
       maintainer="HPE Storage" \
       vendor="HPE" \
-      version="2.0.0" \
+      version="2.1.1.0" \
```
datamattsson, line 10: Is this a new convention with four positions?

datamattsson (approved):
LGTM with questions.

cmd/csi-driver/Dockerfile
raunakkumar: Yes, agreed internally with the team unless it brings up any issues. We will know more once we work on the beta helm chart

datamattsson: If the 4th position is going to be package version, is the container considered the package in this instance then? I think in this case the "version" will be the visible "application version" in the partner ecosystems.

raunakkumar: So should we go with application version 2.1.1 and package version 2.1.1.0 ?

datamattsson:
That would make more sense to me.

PR #314, opened 2022-05-10
Decrease NoExecute toleration on NFS pods

ayush1123:
Signed-off-by: bhagyashree_sarawate <bhagyashree.sarawate@hpe.com>

### DIFF
--- a/pkg/flavor/kubernetes/nfs.go
+++ b/pkg/flavor/kubernetes/nfs.go
@@ -763,6 +763,22 @@ func (flavor *Flavor) makeNFSDeployment(name string, nfsSpec *NFSSpec, nfsNamesp
 
 	volumes = append(volumes, configMapVol)
 
+        var seconds int64 = 30
+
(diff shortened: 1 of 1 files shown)


pkg/flavor/kubernetes/nfs.go
```diff
@@ -763,6 +763,22 @@ func (flavor *Flavor) makeNFSDeployment(name string, nfsSpec *NFSSpec, nfsNamesp
 
 	volumes = append(volumes, configMapVol)
 
+        var seconds int64 = 30
```
datamattsson, line 766: Would it be a huge effort to implement this as a `nfsTolerationSeconds` parameter in the `StorageClass`?

ayush1123: As discussed, we will be taking this as part of another story because we need to change the code and test all over again which would cause a delay in handing over the build to QA for testing.

PR #327, opened 2022-12-03
Fix scrubber going too deep into PV data

gbarazer:
Fix csi-driver scrubber task walking the whole PV filesystems ($KUBELETDIR/pods) where it only needs to find ephemeral_data.json files at a specific depth ($KUBELETDIR/pods/$POD/volumes/kubernetes.io~csi/$PVC/ephemeral_data.json). This causes lots of problems when pods are using PVs containing lots of files/dirs and is even worse when the PV is using another CSI such as NFS.

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -757,6 +757,9 @@ func (driver *Driver) ScrubEphemeralPods(podsDirPath string) error {
 	// Fetch all ephemeral pods and populate it
 	//ephemeralPods := map[string][]*StagingDeviceEphemeralData{}
 	ephemeralPods := map[string][]*VolumeHandleTargetPath{}
+
+	baseDepth := strings.Count(podsDirPath, string(os.PathSeparator));
+
 	err = filepath.Walk(podsDirPath, func(fileFullPath string, info os.FileInfo, walkErr error) error {
 		if walkErr != nil {
 			log.Errorf("Error while processing the path [%s], %s", fileFullPath, walkErr.Error())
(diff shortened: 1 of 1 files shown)


suneeth51 (approved):
lgtm

pkg/driver/driver.go
```diff
@@ -767,8 +770,20 @@ func (driver *Driver) ScrubEphemeralPods(podsDirPath string) error {
 			return filepath.SkipDir /* Skip processing current directory and continue processing other directories */
 		}
 
+		log.Tracef("Walking path [%s]", fileFullPath)
+		
+		if info.IsDir() && (strings.Count(fileFullPath, string(os.PathSeparator)) - baseDepth) == 2 && info.Name() != "volumes" {
```
raunakkumar, line 775: Could you please add a comment here why the difference needs to be 2 for others reading this code.?

pkg/driver/driver.go
```diff
@@ -767,8 +770,20 @@ func (driver *Driver) ScrubEphemeralPods(podsDirPath string) error {
...
+			log.Tracef("Skipping non-volumes path [%s]", fileFullPath)
+			return filepath.SkipDir
+		}
+
+		if info.IsDir() && (strings.Count(fileFullPath, string(os.PathSeparator)) - baseDepth) == 5 {
```
raunakkumar, line 780: Could you also add a comment here as well?

raunakkumar (approved):
Overall the PR looks fine to me. Could you please share the numbers with/without this fix so that we can internally do some test benchmark?

pkg/driver/driver.go
gbarazer: Done

pkg/driver/driver.go
gbarazer: Done

gbarazer:
Added comments for clarity of the path depth logic for skipping files, which is unoptimal due to the use of filepath.Walk.

Regarding numbers, I don't have benchmark reports but we have several clusters using the driver and also using volumes provisioned by the NFS CSI driver. Some of these volumes have millions of files stored in thousands of sharded directories. 

We first noticed : 

1.  our NFS array processing dozens of thousands of metadata IOPS since the moment we deployed the HPE CSI driver
2. the metadata traffic originated from the nodes on which some pods were stuck in terminating state
3. the pods stuck in terminating state were the ones using PVC with lots of files/dirs
4. other nodes and pods were acting normally
5. the Scrubber logs in the csi-driver container indicated oddly long times for the scrubbing process (we are not using ephemeral volumes)

We first increased the delay between scrub tasks to 24 hours, and confirmed that the heavy traffic was related to the scrubbing process, with metadata-heavy IOPS during 2-3 hours.

With this fix, we can't even notice the metadata traffic overhead from the scrubber with 10s resolution metrics because the task completes in less than a second.

raunakkumar:
> 1. of metadata IOPS since the moment we deploye

Thanks for the details. We do need to ensure scrubbing for ephemeral is not broken but we will take that up internally. We should be good to merge this PR

datamattsson:
The driver passes CSI e2e tests with this patch on Rocky 8 with K8s 1.26 and the scrubber reaps volumes as intended.

```
Ran 69 of 7345 Specs in 12635.309 seconds
SUCCESS! -- 69 Passed | 0 Failed | 0 Pending | 7276 Skipped
PASS
```

PR #328, opened 2022-12-05
2.3.0-beta build

datamattsson:


cmd/csi-driver/Dockerfile
```diff
@@ -1,20 +1,20 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.0.0
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.1.0-1656.1669627757
```
raunakkumar, line 1: Are we going with a specific tag or a 9.1.0 latest would suffice?

raunakkumar (approved):

PR #329, opened 2022-12-06
Additional vendoring

datamattsson:


raunakkumar (approved):
lgtm.  Do you need to cut a branch/tag from common-host-libs ?

datamattsson:
There's some issue when pulling in common-host-libs and k8s-custom-resources, those can be manually fixed before building though but we should cut new tags on those repos I assume.

raunakkumar:
> There's some issue when pulling in common-host-libs and k8s-custom-resources, those can be manually fixed before building though but we should cut new tags on those repos I assume.

Sure. Common host libs still doesn't use go.mod but an older vendoring mechanism  glide (prior to go 1.12 when go.mod became default)
We had intern convert it to go.mod https://github.com/hpe-storage/common-host-libs/pull/164 and that maybe needed if we need parity.

PR #356, opened 2023-09-05
Update version + typo

datamattsson:
Signed-off-by: Michael Mattsson michael.mattsson@gmail.com

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN ln -s /usr/lib64/libgcrypt.so.11.8.2 /usr/lib64/libgcrypt.so.11
 LABEL name="HPE CSI Driver for Kubernetes" \
       maintainer="HPE Storage" \
       vendor="HPE" \
-      version="2.4.0-beta" \
+      version="2.4.0-beta2" \
       summary="HPE CSI Driver for Kubernetes" \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN ln -s /usr/lib64/libgcrypt.so.11.8.2 /usr/lib64/libgcrypt.so.11
 LABEL name="HPE CSI Driver for Kubernetes" \
       maintainer="HPE Storage" \
       vendor="HPE" \
-      version="2.4.0-beta" \
+      version="2.4.0-beta2" \
```
dileepds, line 35: do we need to name beta2 ?

cmd/csi-driver/conform/entrypoint.sh
```diff
@@ -12,7 +12,7 @@ done
 
 # Set up node configurations based on environment variables
 disableNodeConformance=${DISABLE_NODE_CONFORMANCE}
-disableNodeConfiguraton=${DISABLE_NODE_CONFIGURATION}
+disableNodeConfiguration=${DISABLE_NODE_CONFIGURATION}
```
dileepds, line 15: not able to catch the type here

Dockerfile
datamattsson: If we decide to push another chart we need new tags.

PR #362, opened 2023-10-18
CON-2508- Stale entry check before starting the CSI driver

AnushaY1916:
- Added a new flag called "DIABLE_NODE_PANIC"
- Stale entry check using rescan every time the CSI driver is started

### DIFF
--- a/cmd/csi-driver/csi-driver.go
+++ b/cmd/csi-driver/csi-driver.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"syscall"
-	"strings"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


cmd/csi-driver/csi-driver.go
```diff
@@ -51,6 +52,18 @@ var (
 			} else {
 				log.InitLogging(csiControllerLogFile, nil, true)
 			}
+			// Scanning for stale entries on each node where CSI resides, 
+			// this check makes sure that it runs only on hpe-csi-nodes pods 
+			podName := os.Getenv("POD_NAME")
+			if strings.Contains(podName, "hpe-csi-node-") { //&& !strings.Contains(podName, "hpe-csi-controller-") {
```
datamattsson, line 58: You should be able to check the arguments if the driver is running in node or controller mode (`--node-service`)

cmd/csi-driver/csi-driver.go
```diff
@@ -51,6 +52,18 @@ var (
...
+        	                log.Info("********Cleaning the stale entries before starting the HPE CSI driver********")
+				log.Info("Running rescan-scsi-bus.sh script to flush the stale multipath devices...")
+				output, _, err := util.ExecCommandOutputWithTimeout("rescan-scsi-bus.sh", []string{"-f", "-m", "-r"}, 120)
+				if err != nil {
+					log.Error("unable to flush the stale multipath devices using rescan-scsi-bus.sh script")
+				}
```
datamattsson, line 64: If this fails, we need to understand why and determine if it's fatal. And if it succeeds, we need to ensure the node is still not unhealthy. If any of those two conditions are met, reboot.

datamattsson (changes requested):

cmd/csi-driver/csi-driver.go
AnushaY1916: Updated

AnushaY1916:
https://github.com/hpe-storage/co-deployments/pull/352 

DISABLE_NODE_PANIC flag handling in co-deployments.

PR #363, opened 2023-10-25
Fix: Device Info file missing

AnushaY1916:


datamattsson (changes requested):
See internal feedback.

AnushaY1916:
Not required.

PR #383, opened 2024-03-05
Force delete fix

datamattsson:
Image that behaves like CSI driver 2.2.0: quay.io/datamattsson/csi-driver:v2.4.1-forcedelete


dileepds (approved):
LGTM

PR #392, opened 2024-03-21
CON-2508: Added Node Monitor and init container

AnushaY1916:
Added node monitor and init container with basic functionality of getting multipath devices and showing whether they are healthy or not

### DIFF
--- a/cmd/csi-driver/csi-driver.go
+++ b/cmd/csi-driver/csi-driver.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/hpe-storage/csi-driver/pkg/driver"
 	"github.com/hpe-storage/csi-driver/pkg/flavor"
+	"github.com/hpe-storage/csi-driver/pkg/initcontainer"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


pkg/initcontainer/initcontainer.go
```diff
@@ -0,0 +1,64 @@
+package initcontainer
```
datamattsson, line 1: I'm questioning the name of this package. An "initContainer" is a Kubernetes pod construct and I think this is confusing me already. I think `nodeinit` would be a more prescriptive name.

pkg/initcontainer/initcontainer.go
```diff
@@ -0,0 +1,64 @@
...
+			if device.IsUnhealthy {
+				log.Infof("Multipath device %s is unhealthy and is present on %s node", device.Name, ic.nodeName)
+			} else {
+				log.Infof("Multipath device %s is healthy and is present on %s node", device.Name, ic.nodeName)
+			}
+		}
+	} else {
+		log.Tracef("No multipath devices found on the node %s", ic.nodeName)
+	}
+	return nil
```
datamattsson, line 63: Why is this all duplicate code of the nodemonitor.go? Is there no way to have a common code base for the monitor and init logic? My initial assumption was that we can call the node monitor thread once during init and then exit? Can we just put all this in a function and call it nodepatrol or something and that will hold all logic for an iteration?

dileepds (approved):
LGTM

datamattsson (commented):
LGTM but I rather have Dileep and Sijeesh do the final review.

cmd/csi-driver/csi-driver.go
```diff
@@ -116,6 +124,16 @@ func csiCliHandler(cmd *cobra.Command) error {
...
 	}
 
+	if nodeInit {
+		nodeInitContainer := nodeinit.NewNodeInitContainer(flavorName, nodeService)
+		err := nodeInitContainer.NodeInit()
+		if err != nil {
+			log.Errorf("Error while running the init container logic: %s", err.Error())
+			os.Exit(1)
+		}
+		os.Exit(0)
```
sijeesh, line 134: exiting here ?

sijeesh (approved):
LGTM

PR #397, opened 2024-04-10
NFS updates for v2.5.0

datamattsson:
- CON-3064
- CON-3084
- CON-3086

### DIFF
--- a/pkg/flavor/kubernetes/nfs.go
+++ b/pkg/flavor/kubernetes/nfs.go
@@ -1136,14 +1136,18 @@ func int32toPtr(i int32) *int32 {
 
 func (flavor *Flavor) getResourceQuantity(scParams map[string]string, paramKey string, defaultVal string) (resource.Quantity, error) {
 
+	var quantityVal string
+
 	if val, ok := scParams[paramKey]; ok {
-		defaultVal = val
+		quantityVal = val
+	} else {
+		quantityVal = defaultVal
 	}
 
-	quantity, err := resource.ParseQuantity(defaultVal)
+	quantity, err := resource.ParseQuantity(quantityVal)
 
 	if err != nil {
-		return quantity, fmt.Errorf("Invalid '%s' value of '%s' provided in Deployment, %s", paramKey, defaultVal, err.Error())
+		return quantity, fmt.Errorf("Invalid '%s' value of '%s' provided in Deployment, %s", paramKey, quantityVal, err.Error())
 	}
 
 	return quantity, nil
--- a/pkg/flavor/kubernetes/nfs.go
+++ b/pkg/flavor/kubernetes/nfs.go
@@ -22,40 +22,46 @@ import (
 )
 
 const (
-	nfsPrefix           = "hpe-nfs-"
-	pvcPrefix           = "pvc-" // this is hardcoded at csi-provisioner runtime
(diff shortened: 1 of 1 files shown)


dileepds (approved):
LGTM

AnushaY1916 (approved):

pkg/flavor/kubernetes/nfs.go
```diff
@@ -453,44 +459,92 @@ func (flavor *Flavor) getNFSSpec(scParams map[string]string) (*NFSSpec, error) {
...
+	resourceRequests[core_v1.ResourceMemory] = memoryRequestQuantity
+
+	// User provided NFS cpu requests eg: cpu=500m
+	if val, ok := scParams[nfsResourceRequestsCPUKey]; ok {
+		quantity, err := resource.ParseQuantity(val)
+		if err != nil {
+			return nil, fmt.Errorf("Invalid '%s: %s' provided in StorageClass, err %s", nfsResourceRequestsCPUKey, val, err.Error())
+		}
+		resourceRequests[core_v1.ResourceCPU] = quantity
+	}
```
sijeesh, line 522: Can we use the 'else' condition here to assign default value only when a user-provided value is absent, thereby avoiding redundant variable assignments?

pkg/flavor/kubernetes/nfs.go
```diff
@@ -453,44 +459,92 @@ func (flavor *Flavor) getNFSSpec(scParams map[string]string) (*NFSSpec, error) {
...
+
+	// User provided NFS memory requests eg: memory=64Mi
+	if val, ok := scParams[nfsResourceRequestsMemoryKey]; ok {
+		quantity, err := resource.ParseQuantity(val)
+		if err != nil {
+			return nil, fmt.Errorf("Invalid '%s: %s' provided in StorageClass, err %s", nfsResourceRequestsMemoryKey, val, err.Error())
+		}
+		resourceRequests[core_v1.ResourceMemory] = quantity
+	}
```
sijeesh, line 531: same as above ?

sijeesh (changes requested):

pkg/flavor/kubernetes/nfs.go
datamattsson: Thanks for calling out the sloppy code. I removed a lot of the duplicate code by replacing it with a function.

pkg/flavor/kubernetes/nfs.go
```diff
@@ -1161,3 +1133,18 @@ func (flavor *Flavor) waitForDeployment(deploymentName string, nfsNamespace stri
 func int32toPtr(i int32) *int32 {
 	return &i
 }
+
+func (flavor *Flavor) getResourceQuantity(scParams map[string]string, paramKey string, defaultVal string) (resource.Quantity, error) {
+
+	if val, ok := scParams[paramKey]; ok {
```
sijeesh, line 1139: Suggestion:
if you think, this can increase the readability

if val, ok := scParams[paramKey]; ok {
		quantityVal = val // User defined
} else {
                quantityVal = defaultVal //Default 
}

sijeesh (approved):
LGTM

pkg/flavor/kubernetes/nfs.go
datamattsson: Thanks, fixed!

PR #410, opened 2024-06-24
CON2508: Cleanup of stale multipath devices for node monitor as well as node init container

AnushaY1916:
The following functionalities are implemented in phase 3:

1. Unmount all references to the device, bind mounts, and real mounts.
2. Kill processes using the mount points and devices
3. Remove block devices
4. Flush the multipath map
5. Enable and disable of node monitor and cleanup in the node init container based on the user's choice

made improvements for the previous  phase's implementation as well

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4-949.1714662671
 RUN microdnf update -y && rm -rf /var/cache/yum
 ADD cmd/csi-driver/AlmaLinux-Base.repo /etc/yum.repos.d/
 
-RUN microdnf install -y cryptsetup tar procps psmisc
+RUN microdnf install -y cryptsetup tar procps
 
 COPY --from=centos:centos7.9.2009 /usr/bin/systemctl /usr/bin/systemctl
 COPY --from=centos:7.9.2009 /usr/lib64/libgcrypt.so.11.8.2 /usr/lib64/libgcrypt.so.11.8.2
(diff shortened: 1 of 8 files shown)


dileepds (approved):
LGTM

Dockerfile
```diff
@@ -22,22 +22,22 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4-949.1714662671
...
-      io.k8s.description="The HPE CSI Driver for Kubernetes enables container orchestrators, such as Kubernetes and OpenShift, to manage the life-cycle of persistent storage." \
-      io.openshift.tags=hpe,csi,hpe-csi-driver
+    maintainer="HPE Storage" \
+    vendor="HPE" \
+    version="2.5.0-beta" \
+    summary="HPE CSI Driver for Kubernetes" \
+    description="The HPE CSI Driver for Kubernetes enables container orchestrators, such as Kubernetes and OpenShift, to manage the life-cycle of persistent storage." \
+    io.k8s.display-name="HPE CSI Driver for Kubernetes" \
+    io.k8s.description="The HPE CSI Driver for Kubernetes enables container orchestrators, such as Kubernetes and OpenShift, to manage the life-cycle of persistent storage." \
+    io.openshift.tags=hpe,csi,hpe-csi-driver
```
datamattsson, line 40: Why change the indent here?

Dockerfile
```diff
@@ -22,22 +22,22 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4-949.1714662671
 RUN microdnf update -y && rm -rf /var/cache/yum
 ADD cmd/csi-driver/AlmaLinux-Base.repo /etc/yum.repos.d/
 
-RUN microdnf install -y cryptsetup tar procps
+RUN microdnf install -y cryptsetup tar procps psmisc
```
datamattsson, line 25: Shouldn't we borrow `fuser` from the host? That's what I did.

datamattsson (changes requested):

Dockerfile
AnushaY1916: The IDE updated it when saved.

Dockerfile
AnushaY1916: Updated

datamattsson (approved):

PR #411, opened 2024-07-09
Handling the ISCSI and Multipath configuration in node init container

AnushaY1916:


datamattsson (changes requested):
Can we fix L132-133 in this PR to read better? This is what the log entry looks like:

```
time="2024-07-09T23:25:35Z" level=info msg="Node configuration is disabled, DISABLE_NODE_CONFIGURATION=true.Skipping the Multipath and ISCSI configurations" file="csi-driver.go:132"
```

Something like this:
```
				" Skipping node configuration. All block storage services needs to be installed and configured manually.", disableNodeConfiguraton)
```

dileepds (approved):
LGTM

sijeesh (approved):
LGTM

datamattsson (approved):

PR #425, opened 2024-12-02
CON-2793 Configure NFS toleration seconds from Storage class

diganta-001:
…s applicable only with nfs volume i.e nfsResources is "true"

### DIFF
--- a/pkg/flavor/kubernetes/nfs.go
+++ b/pkg/flavor/kubernetes/nfs.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -65,6 +66,8 @@ const (
 	nfsProvisionedFromKey        = "provisioned-from"
 	nfsForeignStorageClassKey    = "nfsForeignStorageClass"
 	nfsResourcesKey              = "nfsResources"
+	nfsTolerationSecScKey        = "nfsTolerationSeconds"
+	defaultNfsTolerationSeconds  = 30
 )
 
 // NFSSpec for creating NFS resources
@@ -77,6 +80,7 @@ type NFSSpec struct {
 	labelValue           string
 	sourceNamespace      string
 	sourceVolumeClaim    string
+	tolerationSeconds    *int64
 }
 
 // CreateNFSVolume creates nfs volume abstracting underlying nfs pvc, deployment and service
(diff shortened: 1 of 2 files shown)


datamattsson:
Thank you for the contribution. HPE will need you to perform the DCO sign-off for us to test and merge this feature.

rohit-balakrishna:
@diganta-001 looks like there are some conflicts. Could you please rebase to the latest master and push a new commit ?

pkg/flavor/kubernetes/nfs.go
```diff
@@ -65,6 +66,7 @@ const (
 	nfsProvisionedFromKey        = "provisioned-from"
 	nfsForeignStorageClassKey    = "nfsForeignStorageClass"
 	nfsResourcesKey              = "nfsResources"
+	nfsTolerationSeconds         = "nfsTolerationSeconds"
```
rohit-balakrishna, line 69: This attribute/parameter needs to be defined in a StorageClass right? Should we then name is as `nfsTolerationSecondsStorageClassKey` ?

pkg/flavor/kubernetes/nfs.go
```diff
@@ -861,6 +872,9 @@ func (flavor *Flavor) makeNFSDeployment(name string, nfsSpec *NFSSpec, nfsNamesp
 	}
 
 	var seconds int64 = 30
```
rohit-balakrishna, line 874: we can define a constant for the value 30 instead of hardcoding here.

pkg/flavor/kubernetes/nfs.go
```diff
@@ -556,6 +559,14 @@ func (flavor *Flavor) getNFSSpec(scParams map[string]string) (*NFSSpec, error) {
 	if pvc, ok := scParams[nfsSourcePVCNameKey]; ok {
 		nfsSpec.sourceVolumeClaim = pvc
 	}
+	if tolerationSeconds, ok := scParams[nfsTolerationSeconds]; ok {
+		if num, err := strconv.Atoi(tolerationSeconds); err == nil {
+			num64 := int64(num)
+			nfsSpec.tolerationSeconds = &num64
+		} else {
+			return nil, err
```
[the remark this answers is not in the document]
diganta-001: updated with error format

pkg/flavor/kubernetes/nfs.go
diganta-001: updated the variable name as nfsTolerationSecScKey, suggested variable name nfsTolerationSecondsStorageClassKey
has a indentation issue with previous lines , I can update the variable name to nfsTolerationSecondsStorageClassKey if required

pkg/flavor/kubernetes/nfs.go
diganta-001: added this default value to a constant

rohit-balakrishna:
@diganta-001 its better to update the PR title to include the Jira ID. Same can be the case for the commit message as well.

dileepds (approved):
LGTM

PR #432, opened 2025-01-08
CON-3168: Enable code coverage measurement in csi-driver repository

rohit-balakrishna:
Logged the defect: https://jira.storage.hpecorp.net/browse/CON-3170 to fix the csi sanity unit-test failure. We can merge this PR to enable the unit-test run and coverage measurement.

### DIFF
--- a/pkg/driver/driver_test.go
+++ b/pkg/driver/driver_test.go
@@ -2,7 +2,7 @@
 package driver
 
 import (
-	"os"
+	//	"os"
 	"strconv"
 	"sync"
 	"testing"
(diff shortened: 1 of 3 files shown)


pkg/driver/driver_test.go
```diff
@@ -28,39 +28,40 @@ const (
...
-		TargetPath:      targetPath,
-		Address:         endpoint,
-		SecretsFile:     secretsFile,
-		CreateTargetDir: createTarget,
-	}
-
-	sanity.Test(t, config)
-}
-
+*/
```
dileepds, line 64: Why the sanity testing removed ? I think this will be good for the CSI specification validation.

rohit-balakrishna: @dileepds , the sanity UT was failing. I have created a story https://jira.storage.hpecorp.net/browse/CON-3170 for fixing this failure. Shrey is working on it this sprint. In the meantime, we will enable coverage, since with the sanity commented, UT completes and we have 8.2% coverage. With the check in the Makefile, we will ensure that future merges dont reduce this below 8%.

dileepds: oh ok Thanks.

dileepds (approved):
LGTM

PR #435, opened 2025-02-04
CON-3170:Fix the failures seen in CSI Sanity unit-test case execution

shrey-deshmukh-hpe:
After running multiple test cases, I came to the conclusion that rewriting the old Segments map would have the same effect as commenting out those lines and hence I commented out those lines


shrey-deshmukh-hpe:
Wrong Branch Created

PR #437, opened 2025-02-05
CON-3166 - Removed rollback for NFS resources when creating NFS volume

jlothe:
Please review the rollback removal changes when creating NFS volume for 2.5.2 branch

PR #439, opened 2025-02-17
CON-3113 - Implementation of probes for NFS deployments

AnushaY1916:


dileepds (approved):
LGTM

pkg/flavor/kubernetes/nfs.go
```diff
@@ -152,6 +153,13 @@ func (flavor *Flavor) CreateNFSVolume(pvName string, reqVolSize int64, parameter
 		return nil, true, err
 	}
 
+	log.Tracef("Create a role and role binding for the service account %s", nfsServiceAccount)
+	err = flavor.createRoleAndRoleBinding(nfsServiceAccount, nfsResourceNamespace)
+	if err != nil {
+		log.Errorf("Error occured while creating the role and rolebinding for the ServiceAccount %s:%s", nfsServiceAccount, err.Error())
+		return nil, true, fmt.Errorf("Error occured while creating the role and rolebinding for the ServiceAccount %s:%s", nfsServiceAccount, err.Error())
```
rohit-balakrishna, line 160: Linter generates an error for capitalized error strings. Please change the capitalization for `Error` and `ServiceAccount`

pkg/flavor/kubernetes/nfs.go
```diff
@@ -907,6 +979,44 @@ func (flavor *Flavor) makeNFSDeployment(name string, nfsSpec *NFSSpec, nfsNamesp
...
+
+	livenessProbe := &core_v1.Probe{
+		ProbeHandler: core_v1.ProbeHandler{
+			Exec: &core_v1.ExecAction{
+				Command: []string{"/bin/sh", "/nfsHealthCheck.sh", "3", name, nfsNamespace},
+			},
+		},
+		InitialDelaySeconds: 10,
+		PeriodSeconds:       5,
+		TimeoutSeconds:      4,
```
rohit-balakrishna, line 1012: since the same constants are being used, linter will ask for these to be constants. We can defined global constants for these. Also would we need these parameters to be configurable as part of the StorageClass?

AnushaY1916: I will update the constants. for storage class parameters, please refer the offline discussion

rohit-balakrishna: Makes sense Anusha. thanks.

rohit-balakrishna (approved):

dileepds (approved):

datamattsson (changes requested):
Are we leaving residual `serviceAccounts`, `roles` and `roleBindings` around? I can't see these being pruned in deleteNFSResources.

pkg/flavor/kubernetes/nfs.go
```diff
@@ -907,6 +983,44 @@ func (flavor *Flavor) makeNFSDeployment(name string, nfsSpec *NFSSpec, nfsNamesp
...
+		},
+		InitialDelaySeconds: nfsProbeInitialDelaySeconds,
+		PeriodSeconds:       nfsProbePeriodSeconds,
+		TimeoutSeconds:      nfsProbeTimeoutSeconds,
+	}
+
+	livenessProbe := &core_v1.Probe{
+		ProbeHandler: core_v1.ProbeHandler{
+			Exec: &core_v1.ExecAction{
+				Command: []string{"/bin/sh", "/nfsHealthCheck.sh", "3", name, nfsNamespace},
```
datamattsson, line 1011: Can we maintain a hash of actions for the 1, 2 and 3 in the script instead? I find this numbering very counter-intuitive. nfsNamespace is not being used in the script and should be removed.

datamattsson (changes requested):

datamattsson (approved):
LGTM, can we please work #437 next?

PR #452, opened 2025-05-21
CON-3153 Support for Unified File in R5

diganta-001:
CON-3153 Support for Unified File in R5

CON-3240 Implement the csi-driver changes in CreateVolume to invoke the CSP.

CON-3217 Mount NFS volume to K8s cluster node

CON-3260 Add support for ExpandVolume from the csi-driver

CON-3305 POD Publish does not consider ArrayAccessIP change.


datamattsson (approved):
No e2e regressions with Primera block or NFS server provisioner.

dileepds (approved):

PR #455, opened 2025-06-18
CON-3301 Added changes to skip connection error when creating host in disaster case

jlothe:


pkg/driver/controller_server.go
```diff
@@ -906,10 +906,15 @@ func (driver *Driver) controllerPublishVolume(
 		log.Tracef("Notifying CSP about Node with ID %s and UUID %s", node.ID, node.UUID)
 		log.Tracef("AccessProtocol set is %s", requestedAccessProtocol)
 		node.AccessProtocol = requestedAccessProtocol
-		if err = storageProvider.SetNodeContext(node); err != nil {
-			log.Error("err: ", err.Error())
-			return nil, status.Error(codes.Unavailable,
+		err = storageProvider.SetNodeContext(node)
+		if err != nil {
+			if strings.Contains(err.Error(), "dial tcp") || strings.Contains(err.Error(), "i/o timeout") {
+				log.Errorf("Error connecting to the CSP. err: %s", err.Error())
```
rohit-balakrishna, line 912: We can add a comment stating that we are ignoring the error explicitly and the reasons for the same.

pkg/driver/controller_server.go
```diff
@@ -906,10 +906,20 @@ func (driver *Driver) controllerPublishVolume(
 		log.Tracef("Notifying CSP about Node with ID %s and UUID %s", node.ID, node.UUID)
 		log.Tracef("AccessProtocol set is %s", requestedAccessProtocol)
 		node.AccessProtocol = requestedAccessProtocol
-		if err = storageProvider.SetNodeContext(node); err != nil {
-			log.Error("err: ", err.Error())
-			return nil, status.Error(codes.Unavailable,
-				fmt.Sprintf("Failed to provide node %s to CSP err: %s", node.ID, err.Error()))
+		err = storageProvider.SetNodeContext(node)
+		// Check for error while setting node context. If the error message contains "dial tcp" or "i/o timeout",
+		// these are considered connection errors (e.g., CSP is unreachable or network timeout).
```
rohit-balakrishna, line 911: this should be Array is unreachable right?

rohit-balakrishna (approved):

datamattsson (commented):
I don't quite follow the logic here. Won't this just break very ungraceful in the case of the CSP returns connection errors to the array for any odd reason?

How is this being handled further down the chain during a publish? Unpublish? I thought we'd switch to the secondary array control-plane during a disaster, but that's not the case here?

rohit-balakrishna:
This has been handled in the CSP, closing this PR.

PR #473, opened 2025-11-03
CON-3876 - NVMEoTCP support for csi-driver and host-libs

jlothe:


datamattsson:
Can we make sure the changes needed in hpe-storage-node.sh are part of this PR?

pkg/driver/controller_server.go
```diff
@@ -924,7 +924,9 @@ func (driver *Driver) controllerPublishVolume(
                 requestedAccessProtocol = iscsi
         } else if requestedAccessProtocol == "fc" {
                 requestedAccessProtocol = fc
-        }
+        } else if requestedAccessProtocol == "nvmetcp" {
+		requestedAccessProtocol = "nvmetcp"
```
rohit-balakrishna: looks like this is not correct, line 926, 924 are assigning variables, but here we are assigning the same value, which doesnt look right.

jlothe: in 926 and 924 we are assigning constants and in 928 i was assigning the string directly. Now using constant for it

PR #484, opened 2025-12-11
ESC-16671 Fixed the datasourceRef validation issue for nfs pvc clone

jlothe:
**Root cause:** The PVC clone path updated only spec.dataSource.Name, leaving spec.dataSourceRef unset/mismatched, causing validation/controller confusion and failed.
**Fix:** Explicitly set both DataSourceRef and DataSource to Kind=PersistentVolumeClaim and the child PVC Name, keeping the fields present and consistent across Kubernetes versions.


datamattsson (commented):
I wanted to build and verify this with e2e before approving but master is currently broken.

PR #502, opened 2026-01-30
CON-2683 Duplicate IQNs shouldn't be allowed in hpenodeinfo

rashi-jain1:
JIRA Link : https://jira.storage.hpecorp.net/browse/CON-2683

Bug Description: Duplicate IQNs shouldn't be allowed in hpenodeinfo

Bug Fix: 
In this setup, a Kubernetes cluster was created using three virtual machines:

- One master node (rashi-master)
- One worker node created normally (rashi-worker1)
- One additional worker node (rashi-worker2) created by cloning the rashi-worker1 VM

After cloning, rashi-worker2 was configured and joined to the Kubernetes cluster. As a result, both worker nodes (rashi-worker1 and rashi-worker2) has an identical IQN.

As part of this fix the CSI node driver checks if an IQN is already used by another node before registering. When rashi-worker2 starts, it detects that the IQN already exists for rashi-worker1 and stops the node driver from registering.
Due to this, rashi-worker2 is not added to the HPENodeInfos list.

This behavior is also visible in the logs, where a critical error message is printed indicating that a duplicate IQN was detected and node registration was blocked to prevent data corruption.

### DIFF
--- a/pkg/flavor/kubernetes/flavor.go
+++ b/pkg/flavor/kubernetes/flavor.go
@@ -223,6 +223,15 @@ func (flavor *Flavor) LoadNodeInfo(node *model.Node) (string, error) {
 		}
 		// update node initiator IQNs on mismatch
 		iqnsFromNode := getIqnsFromNode(node)
+		wwpnsFromNode := getWwpnsFromNode(node)
+		nqnsFromNode := getNqnsFromNode(node)
+
+		if err := flavor.validateUniqueInitiators(node.Name, iqnsFromNode, wwpnsFromNode, nqnsFromNode); err != nil {
+			log.Errorf("Initiator validation failed: %s", err.Error())
+			return "", err
+		}
+
+		// update node initiator IQNs on mismatch
 		if !reflect.DeepEqual(nodeInfo.Spec.IQNs, iqnsFromNode) {
 			nodeInfo.Spec.IQNs = iqnsFromNode
 			updateNodeRequired = true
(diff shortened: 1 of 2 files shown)


pkg/flavor/kubernetes/flavor.go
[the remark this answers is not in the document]
rashi-jain1: Done

pkg/flavor/kubernetes/flavor.go
```diff
@@ -285,6 +302,75 @@ func (flavor *Flavor) LoadNodeInfo(node *model.Node) (string, error) {
...
+            continue // Skip self
+        }
+        
+        // Check for duplicate IQNs
+        for _, currentIQN := range iqns {
+            if currentIQN == "" {
+                continue // Skip empty
+            }
+            for _, existingIQN := range existingNodeInfo.Spec.IQNs {
+                if currentIQN == existingIQN {
```
rohit-balakrishna: Should we ignore the case here and compare?

pkg/flavor/kubernetes/flavor_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"fmt"
+	"testing"
+
+	crd_v1 "github.com/hpe-storage/k8s-custom-resources/pkg/apis/hpestorage/v1"
+	"github.com/stretchr/testify/assert"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// validateUniqueInitiatorsForTest is a test-only version that takes node list directly
+// This avoids needing to mock the CRD client and tests the core validation logic
```
rohit-balakrishna: is copilot not able to mock the CRD client?

PR #532, opened 2026-04-27
CON-4621 CSI changes to support accessControlList mutation

rohit-balakrishna:


datamattsson:
No other mutable parameters require CSI driver changes. Why is this needed for file ACLs?

diganta-001:
If mutation parameters are part of the volume, no CSI change is required. However, the access control list belongs to the export configuration. Even if we mutate this data, the VolumeAttribute will not be updated, this is for all.  As a result, every new publish call will continue using the initial AccessClientList unless we explicitly read it from the annotation.

diganta-001:
Here is a block volume sample where we are mutating the Description part, volume level it got updated but no change in kubernetes pv/pvc object level
```
s042N cli% showvv -showcols Name,Comment pvc-dec70c2c-ec67-49ea-a3a7-159
Name                            Comment
pvc-dec70c2c-ec67-49ea-a3a7-159 Need more oomph hello all
---------------------------------------------------------
total
s042N cli%

```
```
volumeAttributes:
      accessProtocol: iscsi
      allowMutations: description,limitIops,limitMbps,performancePolicy
      allowOverrides: description,limitIops,limitMbps,performancePolicy
      compression: "false"
      cpg: SSD_r6
      csi.storage.k8s.io/pv/name: pvc-dec70c2c-ec67-49ea-a3a7-159d5a29f717
      csi.storage.k8s.io/pvc/name: pvc-arcus-130-3par
      csi.storage.k8s.io/pvc/namespace: default
      description: Need more oomph
      fsType: xfs
      hostEncryption: "false"
      provisioningType: tpvv
      snapCpg: SSD_r6
      storage.kubernetes.io/csiProvisionerIdentity: 1772527934077-669-csi.hpe.com
```

dileepds (approved):
LGTM

datamattsson:
> Here is a block volume sample where we are mutating the Description part, volume level it got updated but no change in kubernetes pv/pvc object level

I think we're looking at this at the wrong angle.

1. PVC gets created with accessControlList: "192.168.0.0/16". It now sits in volumeAttributes and the on the API resources on the backend.
2. User realizes the CIDR was too wide. Annotates the PVC with csi.hpe.com/accessControlList: "192.168.50.0/24", the volume-mutator edits via the CSP the API resource on the backend.

In what phase is the value of accessControlList needed by hpe-csi-controller and hpe-csi-node? It's an attribute on the backend API server and becomes a property of the NFS export.

**Edit:** What I'm trying to get across here is that a publish/unpublish for file CSPs should be a no-op as file CSPs does not have the concept of a host and a file "share" might be exported out the realm of Kubernetes and should be exported at the time of provisioning with the requested ACL.

rohit-balakrishna:
> No other mutable parameters require CSI driver changes. Why is this needed for file ACLs?



> > Here is a block volume sample where we are mutating the Description part, volume level it got updated but no change in kubernetes pv/pvc object level
> 
> I think we're looking at this at the wrong angle.
> 
> 1. PVC gets created with accessControlList: "192.168.0.0/16". It now sits in volumeAttributes and the on the API resources on the backend.
> 2. User realizes the CIDR was too wide. Annotates the PVC with csi.hpe.com/accessControlList: "192.168.50.0/24", the volume-mutator edits via the CSP the API resource on the backend.
> 
> In what phase is the value of accessControlList needed by hpe-csi-controller and hpe-csi-node? It's an attribute on the backend API server and becomes a property of the NFS export.
> 
> **Edit:** What I'm trying to get across here is that a publish/unpublish for file CSPs should be a no-op as file CSPs does not have the concept of a host and a file "share" might be exported out the realm of Kubernetes and should be exported at the time of provisioning with the requested ACL.

@datamattsson , as part of the CSI/CSP integration PR, we added a PublishFileVolume interface to the CSP and now both the UnifiedFile and the HomeFleet CSP implement that interface.

Could you elaborate what you mean by: "does not have the concept of a host and a file "share" might be exported out the realm of Kubernetes and should be exported at the time of provisioning with the requested ACL."

During CreateVolume, we will use the StorageClass ACL parameter, which will get populated in the VolumeContext. The same will be used from the VolumeContext during the ControllerPublishVolume when creating the export on the backend. If we now mutate, the backend gets updated, but not the VolumeContext. So we will need to take it from the Annotation value, instead of the VolumeContext during a subsequent Publish. We can discuss the same in our sync-up tomorrow morning.

datamattsson:
> During CreateVolume, we will use the StorageClass ACL parameter, which will get populated in the VolumeContext. The same will be used from the VolumeContext during the ControllerPublishVolume when creating the export on the backend. If we now mutate, the backend gets updated, but not the VolumeContext. So we will need to take it from the Annotation value, instead of the VolumeContext during a subsequent Publish. We can discuss the same in our sync-up tomorrow morning.

I was under the impression that volume and export is one and the same for file CSPs but we've separated those, which is fine.

I also misunderstood how volumeContext was accessed, I thought we were smarter and merged csi.hpe.com annotations on the PVs with with the volumeContext to get the actual volumeContext.

So, CSP implementation details shall leak into the CSI driver, yet again.

rohit-balakrishna:
> > During CreateVolume, we will use the StorageClass ACL parameter, which will get populated in the VolumeContext. The same will be used from the VolumeContext during the ControllerPublishVolume when creating the export on the backend. If we now mutate, the backend gets updated, but not the VolumeContext. So we will need to take it from the Annotation value, instead of the VolumeContext during a subsequent Publish. We can discuss the same in our sync-up tomorrow morning.
> 
> I was under the impression that volume and export is one and the same for file CSPs but we've separated those, which is fine.
> 
> I also misunderstood how volumeContext was accessed, I thought we were smarter and merged csi.hpe.com annotations on the PVs with with the volumeContext to get the actual volumeContext.
> 
> So, CSP implementation details shall leak into the CSI driver, yet again.

Michael, the VolumeContext is passed by CO to ControllerPublishVolume gRPC by copying from the VolumeAttributes. Even if we change the VolumeContext, it will not change the VolumeAttributes(since they are immutable). So even for other mutations that we support through Annotations today, the PV Attributes itself are not updated. The mutations update the Backend only. 

With the VolumeAttributesClass support we will be able to do the mutation in the VolumeAttributes, when we support it. Hope this clarifies.

rohit-balakrishna:
We discussed this with Michael and we will go ahead with this for now. Once we add support for VolumeAttributeClass, we will clean these up.

PR #536, opened 2026-05-18
CON-4645: fsRepair param does not validate the input and no error is reported, silently applies any values input

rashi-jain1:
BUG Link : https://jira.storage.hpecorp.net/browse/CON-4645

BUG Description: fsRepair param does not validate the input and no error is reported, silently applies any values input

BUG Fix :
Added a new validateFsRepairParameter() function that rejects any invalid value of fsRepair  that is not exactly "true" or "false" (lowercase only)
Attached Logs on JIRA.


pkg/driver/node_server.go
```diff
@@ -469,6 +469,12 @@ func (driver *Driver) stageVolume(
 	// Get mount info from the request
 	mountInfo := getMountInfo(volumeID, volCap, publishContext, stagingMountPoint)
 
+	// Validate fsRepair value from volumeContext -- catches PVs that may have been
+	// created before this validation existed and have an invalid (non-boolean) value.
+	if fsRepairVal := volumeContext[fsRepairKey]; fsRepairVal != "" && fsRepairVal != trueKey && fsRepairVal != falseKey {
+		return nil, fmt.Errorf("invalid value %q for fsRepair parameter on volume %s, it must be either [true, false] (lowercase only)", fsRepairVal, volumeID)
```
rohit-balakrishna: We need to return the gRPC error code InvalidArgument here. But the issue is that the NodeStageVolume masks all the error codes into codes.Internal when responding to the CO. We should change this function and the calling function and fix it so that NodeStageVolume can pass err and the actual gRPC error code is prepared in the nodeStageVolume function. Please check the pattern used in NodeUnstageVolume/nodeUnstageVolume. 

While you are changing this, please also fix the same in line no: 763, 771, 777

rashi-jain1: Fixed.

pkg/driver/node_server.go
```diff
@@ -469,6 +470,12 @@ func (driver *Driver) stageVolume(
 	// Get mount info from the request
 	mountInfo := getMountInfo(volumeID, volCap, publishContext, stagingMountPoint)
 
+	// Validate fsRepair value from volumeContext -- catches PVs that may have been
+	// created before this validation existed and have an invalid (non-boolean) value.
```
rohit-balakrishna: If there are customers who had configured this incorrectly before and without this check, it was working without throwing errors, adding this check will cause unnecessary  regressions. I dont think, we should have this check.

CC: @dileepds , @datamattsson

pkg/driver/node_server.go
```diff
@@ -754,21 +761,19 @@ func (driver *Driver) nodeUnstageVolume(volumeID string, stagingTargetPath strin
 				stagingDev.EncKeySecretInfo.Namespace)
 			if err != nil {
 				log.Errorln(err.Error())
-				return err
+				return status.Errorf(codes.Internal, "failed to retrieve encryption passphrase for volume: %s", err.Error())
```
rohit-balakrishna, line 757: the issues also exist at line no: 211, 269, 305, 450, 486, 490, 506, 512, 515, 518. Please change these as well. 
Also at line no: 252(`isVolumeStaged` already add a proper error code) no change needed. 
Once you fix the above and add proper error codes in the `stageVolume` function, you should remove the error code masking at line no: 297.

rashi-jain1: Added.

pkg/driver/node_server.go
rashi-jain1: So, shall I remove this check?

rohit-balakrishna: Yes, please remove this check.

pkg/driver/node_server_test.go
```diff
@@ -159,3 +159,212 @@ func TestNodeGetVolumeStats(t *testing.T) {
...
+				}
+				if tc.errSubstring != "" && !contains(err.Error(), tc.errSubstring) {
+					t.Fatalf("Error %q does not contain expected substring %q", err.Error(), tc.errSubstring)
+				}
+			}
+		})
+	}
+}
+
+func contains(s, substr string) bool {
```
rohit-balakrishna: string module already has function for both contains and containsSubstring, why are we adding our own code. please remove these functions and use the inbuilt string functions.

pkg/driver/controller_server.go
```diff
@@ -1896,3 +1902,17 @@ func (driver *Driver) ControllerModifyVolume(
 	// This driver does not support ControllerModifyVolume, return Unimplemented.
 	return nil, status.Error(codes.Unimplemented, "ControllerModifyVolume is not implemented")
 }
+
+// validateFsRepairParameter validates that the fsRepair StorageClass parameter,
+// if specified, is strictly "true" or "false" (lowercase only).
+func (driver *Driver) validateFsRepairParameter(createParameters map[string]string) error {
```
datamattsson: Can we use a generic library for boolean string validation instead?

datamattsson (changes requested):
Besides my comment, can we validate "nfsResources" and "hostEncryption" the same way as they are string booleans too.

pkg/driver/node_server.go
rashi-jain1: Done.

pkg/driver/node_server_test.go
rashi-jain1: Fixed.

pkg/driver/controller_server.go
rashi-jain1: Done. Added validation for "nfsResources" and for "hostEncryption" we're already validating by "validateHostEncryptionParameters"

dileepds (approved):
LGTM

pkg/driver/node_server_test.go
```diff
@@ -159,3 +160,197 @@ func TestNodeGetVolumeStats(t *testing.T) {
...
+			if err != nil && tc.expectErr && tc.errMsg != "" {
+				if !strings.Contains(err.Error(), tc.errMsg) {
+					t.Fatalf("Error message %q does not contain expected %q", err.Error(), tc.errMsg)
+				}
+			}
+		})
+	}
+}
+
+func TestNodeStageVolumeFsRepairValidation(t *testing.T) {
```
rohit-balakrishna: Since you remove the changes from NodeStageVolume, is this UT needed? Can we remove it if not needed?

pkg/driver/node_server.go
```diff
@@ -476,11 +476,11 @@ func (driver *Driver) stageVolume(
 				log.Infof("Attempting to repair the file system issues of the device %s", device.AltFullPathName)
 				err := driver.chapiDriver.RepairFsckFileSystem(volumeID, device)
 				if err != nil {
-					return nil, fmt.Errorf("Repairing the file system for the volume %s failed due to the error: %v", volumeID, err.Error())
+					return nil, status.Errorf(codes.Internal, "Repairing the file system for the volume %s failed due to the error: %v", volumeID, err.Error())
 				}
 				log.Infof("File system is repaired for the volume %s, proceeding to mount the device %s", volumeID, device.AltFullPathName)
 			} else {
-				return nil, fmt.Errorf("Filesystem issues has been detected and will not be repaired for the volume %s as the fsRepair parameter is not set in the StorageClass", volumeID)
+				return nil, status.Errorf(codes.Internal, "Filesystem issues has been detected and will not be repaired for the volume %s as the fsRepair parameter is not set in the StorageClass", volumeID)
```
rohit-balakrishna: nit: It should be "Filesystem issues have been detected"

pkg/driver/node_server.go
```diff
@@ -496,19 +496,19 @@ func (driver *Driver) stageVolume(
...
 				log.Infof("Retrying to mount after successfully reparing the file system of the volume %s", volumeID)
 				mount, err = driver.chapiDriver.MountDevice(device, mountInfo.MountPoint,
 					mountInfo.MountOptions, mountInfo.FilesystemOptions)
 				if err != nil {
-					return nil, fmt.Errorf("Failed to mount device %s again after successful repair: %v", device.AltFullPathName, err.Error())
+					return nil, status.Errorf(codes.Internal, "Failed to mount device %s again after successful repair: %v", device.AltFullPathName, err.Error())
 				}
 			} else {
-				return nil, fmt.Errorf("Filesystem issues has been detected and will not be repaired for the volume %s as the fsRepair parameter is not set in the StorageClass", volumeID)
+				return nil, status.Errorf(codes.Internal, "Filesystem issues has been detected and will not be repaired for the volume %s as the fsRepair parameter is not set in the StorageClass", volumeID)
```
rohit-balakrishna: nit: It should be "Filesystem issues have been detected"

rashi-jain1: Fixed

pkg/driver/node_server.go
rashi-jain1: Fixed.

pkg/driver/node_server_test.go
rashi-jain1: Yes we can remove it. Done

rohit-balakrishna (approved):

pkg/driver/controller_server.go
```diff
@@ -369,6 +369,14 @@ func (driver *Driver) createVolume(
 		createParameters[hostEncryptionKey] = hostEncryption
 	}
 
+	// Validate the fsRepair and nfsResources parameters
+	for _, paramKey := range []string{fsRepairKey, nfsResourcesKey} {
```
datamattsson, line 373: hostEncryption is a boolean too.

pkg/driver/node_server.go
```diff
@@ -293,15 +294,14 @@ func (driver *Driver) nodeStageVolume(
...
-			fmt.Sprintf("Failed to stage volume %s, err: %s", volumeID, err.Error()))
+		return err
 	}
 	log.Tracef("Staged volume %s successfully, StagingDev: %#v", volumeID, stagingDev)
 
 	// Save staged device info in the staging area
 	log.Tracef("NodeStageVolume writing device info %+v to staging target path %s", stagingDev.Device, stagingTargetPath)
 	if stagingDev == nil || stagingDev.Device == nil {
-		return fmt.Errorf("Invalid staging device info. Staging device cannot be nil")
+		return status.Errorf(codes.Internal, "Invalid staging device info. Staging device cannot be nil")
```
datamattsson, line 304: Why is every return changed from here on forward?

rohit-balakrishna: This was an existing issue, we were not returning proper error codes and CO would have received codes.Unknown instead of proper error codes. We fixed that for the NodeStageVolume and NodeUnstageVolume functions.

datamattsson (approved):
Not tested.

PR #539, opened 2026-05-26
CON-4687: Code change to include MDcache for nfs pod to handle symbolic link

souvanickRepo:
Code change to include MDcache for nfs Ganesha pod to handle symbolic link as raised in escalation.
Jira Link: https://jira.storage.hpecorp.net/browse/CON-4687

### DIFF
--- a/pkg/flavor/kubernetes/nfs.go
+++ b/pkg/flavor/kubernetes/nfs.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -41,6 +42,8 @@ const (
 
 	creationInterval               = 60 // 300s with sleep interval of 5s
 	creationDelay                  = 5 * time.Second
+	configMapUpdateRetryCount      = 5
+	configMapUpdateRetryDelay      = 500 * time.Millisecond
 	defaultExportPath              = "/export"
 	nfsResourceLimitsCPUKey        = "nfsResourceLimitsCpuM"
 	nfsResourceRequestsCPUKey      = "nfsResourceRequestsCpuM"
@@ -352,6 +355,13 @@ NFSv4
   UseGetpwnam = true;
   DomainName = "REPLACE_DOMAIN";
 }
+MDCACHE
+{
+  Dir_Chunk = 0;
+  Entries_HWMark = 50000;
+  LRU_Run_Interval = 90;
+  Use_Getattr_Directory_Invalidation = true;
+}
 EXPORT
 {
   Export_Id = 716;
@@ -362,6 +372,7 @@ EXPORT
   Transports = TCP;
   Protocols = 4;
   SecType = "sys";
+  Attr_Expiration_Time = -1;
   FSAL {
       Name = VFS;
   }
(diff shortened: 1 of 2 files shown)


pkg/flavor/kubernetes/nfs_test.go
```diff
@@ -102,6 +102,22 @@ func TestCreateConfigMap(t *testing.T) {
...
+	assert.NotContains(t, ganeshaConfig, "REPLACE_DOMAIN")
+
+	// Validate key NFS and MDCACHE defaults
+	assert.Contains(t, ganeshaConfig, "NFS_Protocols= 4;")
+	assert.Contains(t, ganeshaConfig, "MDCACHE")
+	assert.Contains(t, ganeshaConfig, "Dir_Chunk = 0;")
+	assert.Contains(t, ganeshaConfig, "Entries_HWMark = 50000;")
+	assert.Contains(t, ganeshaConfig, "LRU_Run_Interval = 90;")
+	assert.Contains(t, ganeshaConfig, "Cache_FDs = false;")
+	assert.Contains(t, ganeshaConfig, "Use_Getattr_Directory_Invalidation = true;")
```
rohit-balakrishna, line 131: // Missing assertions:
assert.Contains(t, ganeshaConfig, "fsid_type = UUID;")
assert.Contains(t, ganeshaConfig, "Delegations = false;")
assert.Contains(t, ganeshaConfig, "Attr_Expiration_Time = -1;")

pkg/flavor/kubernetes/nfs.go
```diff
@@ -333,6 +342,8 @@ EXPORT
   Transports = TCP;
   Protocols = 4;
   SecType = "sys";
+  Delegations = false;
+  Attr_Expiration_Time = -1;
```
rohit-balakrishna, line 375: @souvanickRepo , the ConfigMap holding these parameters are created when the first NFS PVC get created in a given namespace. If it exists, we are silently ignoring the updates. Please check the ConfigMap creation path and add code to update it, if the ConfigMap already exists.

rohit-balakrishna (changes requested):

pkg/flavor/kubernetes/nfs.go
kalikivayi: @souvanickRepo I have requested the Copilot to review the changes. The copilot suggested the comment provided by Rohit as well as one more change.
The `Attr_Expiration_Time` should be changed to 0 or its ok even if we omit it altogether.
Please find the complete analysis in the attached .md file.

[CON-4687-Code-Review.md](https://github.com/user-attachments/files/29841094/CON-4687-Code-Review.md)

kalikivayi (changes requested):
Please address the review comments.

pkg/flavor/kubernetes/nfs.go
souvanickRepo: @kalikivayi,
without Attr_Expiration_Time = -1 or any other value, it was not working, conversation and response on co-pilot on the same earlier:
---------------------------------------
Why Working YAML likely works

The combination of MDCACHE hardening plus Attr_Expiration_Time = -1 and Delegations = false reduces stale metadata/handle transitions during readdir and symlink traversal.
Cache_FDs = false is especially strong for avoiding stale FD-driven failure modes

souvanickRepo: @rohit-balakrishna , Added the updated part.

souvanickRepo:
Review from Co-pilot after adding update:
# ESC-17375 Re-Review: NFS ConfigMap Update Logic
Assumption: As we are introducing the update of config map, so expecting no third-party will be updating the same config map.

Date: 2026-08-03
Scope: Review of the patch captured for RWX PVC symlink/mount behavior.

## Context and Accepted Assumptions

1. Ganesha pod restart/re-initialization is guaranteed after ConfigMap update, so config reload is triggered on the cluster.
2. The specific ConfigMap is owned and managed only by CSI controller components.
3. Symlink and RWX mount behavior validation is performed at cluster/integration level (not unit-test level).
4. Duplicate assertion concern was already updated in the patch.

## Re-Review Verdict

No blocking findings under the above assumptions.

The change is acceptable for ESC-17375 based on current code behavior and the stated operational/test guarantees.

## What Went Well / What Fixed the Issue

- The logic changed from create-only to create-or-update for the NFS ConfigMap, ensuring corrected Ganesha config is applied even when the ConfigMap already exists.
- The update path includes conflict-aware retry behavior, which improves reliability when multiple controller actions race on the same object.
- The no-op guard prevents unnecessary updates when the desired Data/Labels already match, reducing churn and avoiding needless restarts.
- With the accepted operational assumption that Ganesha is restarted/re-initialized after ConfigMap change, updated NFS settings are guaranteed to be reloaded on cluster nodes.
- This directly addresses the typical failure mode behind ESC-17375 where corrected NFS behavior would not take effect if an existing ConfigMap were left unchanged.

In short: previously, a stale existing ConfigMap could preserve old NFS behavior; now the controller converges that ConfigMap to the desired state, and cluster restart/reload applies the fix to running RWX workloads.

## What Was Re-Checked

- ConfigMap create/update path handles create races and update conflicts with retry.
- No-op guard avoids unnecessary update calls when Data/Labels are already in desired state.
- Test updates validate:
  - Config content additions for NFS/Ganesha sections.
  - ConfigMap update behavior on second invocation.
  - No-op behavior on third invocation with unchanged desired state.

## Remaining Non-Blocking Risk

- Unit tests do not prove end-to-end symlink traversal semantics through a mounted RWX pod.
- This is acceptable because the behavioral validation is explicitly covered by cluster testing for ESC-17375.

pkg/flavor/kubernetes/nfs_test.go
souvanickRepo: updated

kalikivayi (approved):
LGTM

pkg/flavor/kubernetes/nfs.go
```diff
@@ -362,6 +372,8 @@ EXPORT
...
```
rohit-balakrishna, line 375: With these changes to the ConfigMap, can you post the startup log of the NFS Ganesha Server and ensure that there are no warnings being thrown for the new parameters that we have added. You can: parse with ganesha.nfsd -N NIV_FULL_DEBUG  OR check startup logs for “unknown parameter” errors or warnings. Please post the startup log to the bug or here.

pkg/flavor/kubernetes/nfs.go
```diff
@@ -379,15 +391,58 @@ EXPORT
...
+		if err == nil {
+			log.Debugf("configmap %s successfully updated in namespace %s", nfsConfigMap, nfsNamespace)
+			return nil
+		}
+
+		if !errors.IsConflict(err) {
+			return err
+		}
+
+		time.Sleep(creationDelay)
```
rohit-balakrishna: `creationDelay` and `creationInterval` are used for the overall NFS PVC creation logic. For the ConfigMap update race, this is too big a wait. We can try for 5 times with sub-second (0.5 sec) delay

pkg/flavor/kubernetes/nfs.go
souvanickRepo: [nfs.log](https://github.com/user-attachments/files/30833153/nfs.log)
Co-pilot analysis against the log for parameter related changes:

From this log, no hard ConfigMap parsing/apply failure is visible.

What I see on config-related behavior:

Config parsed successfully: nfs.log:245.
Core blocks are loaded and committed without WARN/CRIT parse errors:
NFS_Core_Param commit: nfs.log:176
NFSv4 commit: nfs.log:211
MDCACHE commit: nfs.log:244
EXPORT created successfully: nfs.log:547
Parameter changes appear to be taking effect:
NFS_Protocols changes from default 3 to loaded value 2: nfs.log:138, nfs.log:174
Graceless is active (server skips grace): nfs.log:208, nfs.log:320

souvanickRepo:
e2e run successfully, i.e. with CSI_E2E_CONFIG=primera-iscsi-multiwriter-ocp:
image

pkg/flavor/kubernetes/nfs.go
rohit-balakrishna: ok, thanks. Looks like you removed the `Delegations = false` variable in the latest commit. I see it was there in the previous commit. Why was that? 
I checked the overall set of config params for consistency and conflicts if any and the `Attr_Expiration_Time = -1;` seems in conflict with the bug that the change intends to fix. 

Please refer to Section 6 of the attached md file.
[CON-4687-nfs-mdcache-symlink-review.md](https://github.com/user-attachments/files/31187503/CON-4687-nfs-mdcache-symlink-review.md)

souvanickRepo: Have to remove the Delegations = false as with this param noticed "unknown param errors" in the logs, also as mentioned earlier without Attr_Expiration_Time = -1; the fix is not working, recent observation from the co-pilot for the same:
------------------------------------
Understood. If removing Attr_Expiration_Time = -1; makes symlink EIO/staleness come back in your environment, then keeping it is the correct behavior-driven choice.

What this means in practice:

Your Ganesha build/runtime likely has semantics where explicit EXPORT-level Attr_Expiration_Time = -1; is required to avoid stale symlink attrs.
So the review assumption about -1 being a no-op or harmful is not universally true for your deployed version.
In your code, this is currently set at nfs.go:376, with directory invalidation already enabled at nfs.go:363.

rohit-balakrishna (approved):

dileepds (approved):

PR #549, opened 2026-06-18
CON-4502: Fixed issues reported by MCP

kalikivayi:
This PR contains Fixes reported by the AI.
Analyzed the report and fixed the issues marked for Fixing in the attached Excel sheet.
[3par-csi-issues-analysis.xlsx](https://github.com/user-attachments/files/29100188/3par-csi-issues-analysis.xlsx)
[3par-csp.md](https://github.com/user-attachments/files/29100190/3par-csp.md)

### DIFF
--- a/pkg/driver/constants.go
+++ b/pkg/driver/constants.go
@@ -2,6 +2,8 @@
 
 package driver
 
+import "time"
+
 const (
 	// Protocol types
 	iscsi           = "iscsi"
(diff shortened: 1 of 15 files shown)


pkg/driver/constants.go
```diff
@@ -53,6 +55,26 @@ const (
 	defaultConnectionMode = "manual"
 	KubeletRootDirEnvKey  = "KUBELET_ROOT_DIR"
 
+	// Replication (Remote Copy) StorageClass parameter keys. These are the
+	// camelCase keys as they appear in a StorageClass/PVC before being converted
+	// to snake_case for the CSP. They are used to detect when array-based
+	// replication is requested so that unsupported protocol combinations (e.g.
+	// NVMe/TCP + replication, CROSS-2) can be rejected early in the driver.
+	remoteCopyGroupKey                    = "remoteCopyGroup"
```
rohit-balakrishna, line 63: I understand the intent of validating the SC parameters as soon as possible, but this should not be done in the CSI Driver. These parameters are best understood by the CSP and it should be validated in the CSP. Bringing that logic here, will violate the current architecture of keeping CSI Driver generic. 
CC: @datamattsson , @dileepds

pkg/driver/utils.go
```diff
@@ -126,12 +129,144 @@ func removeDataFile(dirPath string, fileName string) error {
...
+	}
+	for _, addr := range addresses {
+		if ipFamily(addr) == ipv6Family {
+			return fmt.Errorf("NVMe/TCP only supports IPv4 discovery addresses; got IPv6 address %q", addr)
+		}
+	}
+	return nil
+}
+
+// isReplicationRequested reports whether the supplied StorageClass/PVC create
```
rohit-balakrishna, line 195: CSI Driver is not aware of what Replication is. We should not be doing these validations here. I prefer these to be moved to the CSP.
CC: @datamattsson , @dileepds

datamattsson (changes requested):
Can we break this down to 53 different PRs instead and start with the most critical issue that we're trying fix?

Submit CSI e2e test results from each supported CSP with each PR.

PR #550, opened 2026-06-24
CON-4745: Update the image repositories to use latest UBI images

rohit-balakrishna:


datamattsson (approved):
Assuming the image built in 3.3.0-alpha1 chart.

PR #551, opened 2026-06-25
CON-4768: ESC-17651: NFS POD failed to start resulting in Application PODs outage

rohit-balakrishna:


pkg/flavor/kubernetes/nfs.go
```diff
@@ -1121,6 +1121,7 @@ func (flavor *Flavor) makeNFSDeployment(name string, nfsSpec *NFSSpec, nfsNamesp
 			Volumes:                   volumes,
 			HostIPC:                   false,
 			HostNetwork:               false,
+			EnableServiceLinks:        boolToPtr(false),
```
rohit-balakrishna: @datamattsson, please let me know if we need to make this configurable through StorageClass parameter.

datamattsson: No, this should be false. The NFS Pod does not need access to any Services AFAIK, check with the probes to ensure they don't use the Service name. If so, just set it to `false`.

pkg/flavor/kubernetes/nfs.go
```diff
@@ -1350,6 +1401,14 @@ func int32toPtr(i int32) *int32 {
...
 }
 
+func boolToPtr(b bool) *bool {
+	return &b
+}
+
+func boolToPtr(b bool) *bool {
+	return &b
+}
+
```
datamattsson: Can we remove these and just set L1158 to `false` like the rest?

datamattsson (changes requested):

pkg/flavor/kubernetes/nfs.go
rohit-balakrishna: EnableServiceLinks is a bool pointer:
`EnableServiceLinks *bool `json:"enableServiceLinks,omitempty" ...``
It needs to be ptr, instead of adding custom functions, I am now using ptr.To from the standard package: k8s.io/utils/ptr

datamattsson (approved):
Passes e2e on OpenShift (no snapshots/cloning).

```
Ran 47 of 7673 Specs in 2285.689 seconds
SUCCESS! -- 47 Passed | 0 Failed | 0 Pending | 7626 Skipped
PASS
Created /data/csi-e2e-z2ms9-26-07-02-17-55-14.xml.gz
```

dileepds (approved):

PR #537, opened 2026-05-21
CON-4643 Enable full CSI lifecycle management (publish, expand, delete, snapshot) for statically provisioned and pre-existing array volumes

rashi-jain1:
BUG Link : https://jira.storage.hpecorp.net/browse/CON-4643

Confluence Page Link : https://confluence.storage.hpecorp.net/x/lEqGJw

Attached Logs on JIRA

PR #554, opened 2026-06-26
CON-4710 The NodePublish operation has a dependency on the array to c…

diganta-001:
…heck if the volume is present. However, if we have reached this state, it means the volume should already be present. At scale, removing this check significantly reduces pressure on the CSP.We also observed that at scale, even if the volume was created on the array, the CRD creation sometimes failed silently. In these cases, the PVC becomes Bound, but NodePublish fails because GetVolumeByID depends on the CRD.The CSP CRD issue is being resolved in a separate PR.

### DIFF
--- a/pkg/driver/node_server.go
+++ b/pkg/driver/node_server.go
@@ -971,12 +971,6 @@ func (driver *Driver) nodePublishVolume(
 		volumeID, stagingTargetPath, targetPath)
 	defer log.Trace("<<<<< nodePublishVolume")
 
-	// Get Volume
-	if _, err := driver.GetVolumeByID(volumeID, secrets); err != nil {
-		log.Error("Failed to get volume ", volumeID)
-		return status.Error(codes.NotFound, err.Error())
-	}
-
 	// Read device info from the staging area
 	stagingDev, err := readStagedDeviceInfo(stagingTargetPath)
 	if err != nil {


rohit-balakrishna (approved):

datamattsson:
Has this been tested with all CSPs?

diganta-001:
> Has this been tested with all CSPs?

```
We have verified with Nimble CSP, below is the result 

root@sjjcs:~# kubectl get pods
NAME                         READY   STATUS    RESTARTS   AGE
depnimble-5d8d8b877d-wbjkl   1/1     Running   0          14s
root@sjjcs:~# kubectl get pvc
NAME           STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS     VOLUMEATTRIBUTESCLASS   AGE
nimble-pvc-2   Bound    pvc-19d9dcdd-2733-46cf-a8c5-f1579359b7d3   4Gi        RWO            nimble-sc        <unset>                 11m
Nimble OS $ vol --list | grep pvc-19d9dcdd-2733
pvc-19d9dcdd-2733-46cf-a8c5-f1579359b7d3      4096 Yes    N/A             0         0     100 default:/
Nimble OS $

pod pvc description 
volumes:
  - name: vol1
    persistentVolumeClaim:
      claimName: nimble-pvc-2
```

kalikivayi (approved):
LGTM

datamattsson (approved):
Passes e2e along with #537

PR #561, opened 2026-07-21
Sanitize node hostname before registering it as an array host name

amitosw15:
## Summary

`nodeGetInfo()` (`pkg/driver/node_server.go`) sets `model.Node.Name` directly from `os.Hostname()` with no length validation. `ControllerPublishVolume` sends this value verbatim to the CSP via `SetNodeContext` (`POST /containers/v1/hosts`), which registers it as the host name on the array.

HPE 3PAR/Primera/Alletra arrays cap object names at 31 characters. Some CSPs also prepend their own protocol prefix (e.g. `iqn-`, `nqntcp-`) before forwarding the name to the array, further reducing the usable budget. Kubernetes node hostnames on managed clusters routinely exceed this — e.g. AKS (`aks-nodepool1-31721111-vmss000012`, 34 chars) or EKS/OpenShift-on-AWS (`ip-10-0-135-79.us-east-2.compute.internal`, 42 chars). A node with a hostname over the limit can never register as a host on the array, and therefore can never publish volumes to it.

## Changes

- Add `flavor.SanitizeHostName()` (`pkg/flavor/hostname.go`): hashes the node's FQDN + UUID (not just the short name) into a `csi-<hash>` identifier, so two different nodes — even across tenants sharing naming conventions — can't collide onto the same sanitized value.
- Opt-in via `DISABLE_HOSTNAME` env var, off by default, so existing installs keep their current (unsanitized) host name unless explicitly enabled.
- Once opted in, **every** node's name is hashed, not just ones over the limit — otherwise a cluster ends up with a mix of real and hashed names on the array, an inconsistent naming convention.
- The length limit is configurable via `MAX_HOSTNAME_LEN` (default 31, the platform's real limit) instead of a single hardcoded 31-as-our-own-budget. `SanitizeHostName` reserves room internally for the longest known CSP-added protocol prefix (`nqntcp-` for NVMe/TCP, 7 chars) plus our own `csi-` prefix, so the sanitized name stays safe regardless of which prefix the CSP adds (e.g. `nqntcp-csi-<20 hex chars>` = 31, `iqn-csi-<20 hex chars>` = 28).
- The sanitize call lives in `ControllerPublishVolume` (`pkg/driver/controller_server.go`), right before `SetNodeContext`, not in `nodeGetInfo()`. Sanitizing at node registration time doesn't work for Kubernetes deployments: `pkg/flavor/kubernetes`'s `LoadNodeInfo` overwrites `node.Name` with the real k8s node name (needed elsewhere for `GetNodeLabelsByName` topology lookups), which would silently undo the sanitization. Applying it only at the point the name is actually sent to the array avoids that conflict.
- Log when sanitization occurs, so a sanitized name can be traced back to the real node it came from.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./pkg/...`
- [x] Live-tested against a real OpenShift + HPE Primera cluster with a node hostname (28 chars) that the array was actively rejecting (`invalid input: string length exceeds limit`, code 57). With `DISABLE_HOSTNAME=true`, host registration succeeded and a new PVC/Pod fully attached, mounted, and read/wrote data, tested at both `MAX_HOSTNAME_LEN=25` and `27`.
- [x] Confirmed CSI plugin registration (`GetNodeLabelsByName`) still works correctly — this is what broke when sanitization was first applied at the node-registration layer instead of `ControllerPublishVolume`.
- [x] Verified the sanitized name length against the CSP's own protocol prefix (`iqn-`/`nqntcp-`) matches the platform's 31-character limit exactly.

### DIFF
--- a/pkg/flavor/hostname.go
+++ b/pkg/flavor/hostname.go
@@ -1,0 +1,41 @@
+// Copyright 2026 Hewlett Packard Enterprise Development LP
+
+package flavor
+
+import (
+	"crypto/sha256"
+	"fmt"
+)
+
+const (
+	// DisableHostnameEnvKey opts a node in to registering a hashed hostname with the backend.
+	DisableHostnameEnvKey = "DISABLE_HOSTNAME"
+
+	// MaxHostnameLenEnvKey overrides DefaultMaxHostnameLen when DisableHostnameEnvKey is set.
+	MaxHostnameLenEnvKey = "MAX_HOSTNAME_LEN"
+
+	// DefaultMaxHostnameLen is the platform's naming limit, not our own name's budget.
+	DefaultMaxHostnameLen = 31
+
+	// BackendGeneratedHostnameAnnotationKey stores the backend-generated hostname, separate from ObjectMeta.Name.
+	BackendGeneratedHostnameAnnotationKey = "storage.hpe.com/backend-generated-hostname"
+
+	sanitizedHostnamePrefix = "csi-"
+
+	// cspProtocolPrefixReserve covers the longest CSP-added prefix ("nqntcp-" for NVMe/TCP).
+	cspProtocolPrefixReserve = len("nqntcp-")
+)
+
+// SanitizeHostname always hashes the FQDN+UUID so every opted-in node follows the same
+// naming convention, reserving room in maxLen for the CSP's own protocol prefix.
+func SanitizeHostname(fqdn, uuid string, maxLen int) string {
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(fqdn+uuid)))
+	hashLen := maxLen - cspProtocolPrefixReserve - len(sanitizedHostnamePrefix)
+	if hashLen < 0 {
+		hashLen = 0
+	}
+	if hashLen > len(hash) {
+		hashLen = len(hash)
+	}
+	return sanitizedHostnamePrefix + hash[:hashLen]
+}
--- a/pkg/flavor/kubernetes/flavor.go
+++ b/pkg/flavor/kubernetes/flavor.go
@@ -8,12 +8,14 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/hpe-storage/common-host-libs/chapi"
 	log "github.com/hpe-storage/common-host-libs/logger"
 	"github.com/hpe-storage/common-host-libs/model"
+	hpeflavor "github.com/hpe-storage/csi-driver/pkg/flavor"
 	crd_v1 "github.com/hpe-storage/k8s-custom-resources/pkg/apis/hpestorage/v1"
 	crd_client "github.com/hpe-storage/k8s-custom-resources/pkg/client/clientset/versioned"
 	v1beta1 "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
@@ -189,6 +191,10 @@ func (flavor *Flavor) LoadNodeInfo(node *model.Node) (string, error) {
 	log.Tracef(">>>>> LoadNodeInfo called with node %v", node)
 	defer log.Trace("<<<<< LoadNodeInfo")
 
+	// Capture the (possibly sanitized) backend-facing hostname before it's overwritten below.
+	backendHostname := node.Name
+	hostnameSanitizationEnabled, _ := strconv.ParseBool(os.Getenv(hpeflavor.DisableHostnameEnvKey))
+
 	// overwrite actual host name with node name from k8s to be compliant
 	if name := os.Getenv("NODE_NAME"); name != "" {
 		node.Name = name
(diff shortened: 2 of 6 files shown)


datamattsson (changes requested):
Since this is a breaking change we would need the hashed nodename to be optional that would require a user to opt-in at CSI driver install-time which hostname scheme to use.

Suggestions:
- Pick an environment variable, we [use a pattern](https://artifacthub.io/packages/helm/hpe-storage/hpe-csi-driver#configuration-and-installation) of "disableFeature" which is "false" by default. So setting `DISABLE_HOSTNAME="any value"` would enable hashed nodenames as you suggest in the PR, not setting `DISABLE_HOSTNAME` would use the current behavior.
- The environment variable would be then implemented in the Helm chart and generated with Helm-based Operator, `disableHostname: false` would employ current behavior and `true` would enable hashed hosts. The description of the attribute would be "Set to `true` to generate a unique hash for storage backends that have hostname length limitations".
- The prefix should be "csi-" instead of "k8s-".
- The hash should use the node FQDN + node UUID to prevent multiple tenants using identical patterns for node naming.

amitosw15:
Companion Helm chart / operator change to expose `disableHostname`/`maxHostNameLen` at install time: hpe-storage/co-deployments#468

datamattsson:
The way this behaves right now is undesirable.

If a user pick disableHostname, ALL hostnames, regardless of length, should be turned into a hash. Otherwise you'll end up with conflicting node name conventions for "my-foobar-infrastructure-node-0001" and "my-foobar-compute-node-0001"

The default max length for hashed names should have a ceiling for the "nqntcp-" prefix when using NVMe/TCP nodes. The string "nqntcp-csi-" I count to a default max length of 20 unique characters after the prefix. That said, for the user supplied number, it should be 31 which is the platform limitation we're trying to mitigate, 31 - 11 = 20, i.e. `nqntcp-csi-e4f3235a863002550a03` or `iqn-csi-e4f3235a863002550a03`.

datamattsson (changes requested):
Setting disable hostname and max length to 12 in an iSCSI environment, I would expect the hostname to come out as "iqn-csi-N", instead:

CSI node:
```
$ k logs -nhpe-storage ds/hpe-csi-node |grep "sanitization"
time="2026-07-31T21:31:13Z" level=info msg="Host name sanitization enabled, registering \"threefiddy-cpu2\" as \"csi-3\"" file="node_server.go:2233"
```

CSI controller:
```
k logs -nhpe-storage deploy/hpe-csi-controller |grep "sanitization"
<empty>
```

The host being created on the array: `iqn-threefiddy-cpu2` and the same in `crd/hpevolumeinfos`.

pkg/flavor/hostname.go
```diff
@@ -0,0 +1,41 @@
...
+	DisableHostNameEnvKey = "DISABLE_HOSTNAME"
+
+	// MaxHostNameLenEnvKey overrides DefaultMaxHostNameLen when DisableHostNameEnvKey is set.
+	MaxHostNameLenEnvKey = "MAX_HOSTNAME_LEN"
+
+	// DefaultMaxHostNameLen is the platform's naming limit, not our own name's budget.
+	DefaultMaxHostNameLen = 31
+
+	// ArrayHostNameAnnotationKey stores the array-facing host name, separate from ObjectMeta.Name.
+	ArrayHostNameAnnotationKey = "storage.hpe.com/array-host-name"
```
datamattsson: We refer to an "array" as "backend" in the CSI driver. I suggest the annotation should read `storage.hpe.com/backend-generated-hostname`

In the casing of all variables I just noticed that "HostName" is being used, we refer to hostname in the sense it's being used as part of a domain which is simply "Hostname", without a capitalized N.

pkg/flavor/kubernetes/flavor.go
```diff
@@ -252,6 +256,14 @@ func (flavor *Flavor) LoadNodeInfo(node *model.Node) (string, error) {
 			nodeInfo.Spec.NQNs = nqnsFromNode
 			updateNodeRequired = true
 		}
+		// update the array-facing host name annotation on mismatch
+		if nodeInfo.ObjectMeta.Annotations[hpeflavor.ArrayHostNameAnnotationKey] != arrayHostName {
```
datamattsson: The annotation should only be used if hostname is disabled. Installing the driver with the defaults, the annotation is added as it sits now.

```
items:
- apiVersion: storage.hpe.com/v1
  kind: HPENodeInfo
  metadata:
    annotations:
      storage.hpe.com/array-host-name: threefiddy-cpu1
```

Think of it that these code paths should not be exercised unless disableHostname is set.

pkg/flavor/types.go
```diff
@@ -25,6 +25,7 @@ type Flavor interface {
 	LoadNodeInfo(*model.Node) (string, error)
 	UnloadNodeInfo()
 	GetNodeInfo(nodeID string) (*model.Node, error)
+	GetNodeArrayHostName(nodeID string) (string, error)
```
datamattsson: This naming is ambiguous. `GetNodeBackendHostname` would be a better choice.

datamattsson (changes requested):

datamattsson (approved):
Passes CSI e2e with hostname disabled on and off. When it's off, which is the default, no extra annotations or code paths are being exercised. The implementation is entirely optional, it may be chosen in situations where needed for greenfield deployments.

Installing [the chart](https://github.com/hpe-storage/co-deployments/pull/468) with `--set disableHostname=true --set maxHostnameLength=13` produces the following annotation on the `hpenodeinfos`:

```
...
kind: HPENodeInfo
metadata:
  annotations:
    storage.hpe.com/backend-generated-hostname: csi-15
...
```

This is what shows up as an IQN `hpevolumeinfos` CRD:

```
...
spec:
  record:
    HostName: iqn-csi-15
...
```

The host is also created on the array with this name. 

```
30 iqn-csi-15                     Generic-ALUA iqn.1994-05.com.redhat:a9e4983cb663                                                     0:3:1
                                               iqn.1994-05.com.redhat:a9e4983cb663                                                     0:4:1
                                               iqn.1994-05.com.redhat:a9e4983cb663                                                     1:3:1
                                               iqn.1994-05.com.redhat:a9e4983cb663                                                     1:4:3
```

The generated name is based on "nqntcp-csi-36" which would be the `maxHostnameLength=13` requested by the user. In reality, this would be set to 31 which is the limitation or 3PAR derived platforms which would allow any cloud generated hostname which sometimes include a full UUID along with other identifiers such as role and mnemonic names.

@rohit-balakrishna @dileepds please review.

rohit-balakrishna (approved):

